PR #10, opened 2022-01-14
ARC-272 Implement meemoo IDP

bartnaessens:
Created abstract SamlService with common logic. Extended by both the MeemooService and HetArchiefService

### DIFF
--- a/src/modules/auth/services/meemoo.service.ts
+++ b/src/modules/auth/services/meemoo.service.ts
@@ -2,6 +2,7 @@ import { Injectable } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
 
 import { SamlService } from './saml.service';
+
 @Injectable()
 export class MeemooService extends SamlService {
 	constructor(protected configService: ConfigService) {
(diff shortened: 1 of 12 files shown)


src/modules/auth/services/meemoo.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+
+import { SamlService } from './saml.service';
```
andrych, line 4: Idem, nog een enter

src/modules/auth/services/saml.service.ts
```diff
@@ -0,0 +1,123 @@
+import { Logger } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import got from 'got';
+import { get } from 'lodash';
+import saml2, { IdentityProvider, ServiceProvider } from 'saml2-js';
+import convert = require('xml-js');
```
andrych, line 6: Is die require hier nodig?

PR #27, opened 2022-02-08
feat(arc-278): replace query param object for spaces with primitives

bertyhell:
Since it is a get request all @Query params will be strings. so i think we still need to parse them before using them.
I'll test it out with an object once i add the swagger docs to nestjs.

example of the request:
image


bertyhell:
Already works correctly. NestJS is smarter than i thought :D

PR #30, opened 2022-02-09
feat(arc-294): add media module

bartnaessens:
- search items in elasticsearch adhv query parameter (freetext search)
- paginering
- get by id

Facets / sortering nog uit te werken


src/modules/media/controllers/media.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@Controller('media')
+export class MediaController {
+	private logger: Logger = new Logger(MediaController.name, { timestamp: true });
+
+	constructor(private mediaService: MediaService) {}
+
+	@Post()
+	public async getMedia(@Body() query: MediaQueryDto): Promise<any> {
+		const media = await this.mediaService.findAll(query);
```
bertyhell, line 14: maybe rename query to queryInfo or queryObj so we keep a differentiation between:
* query the object containing filters and paging info
* and query the text a user inputs to search for

bertyhell (approved):

.env/.env.template
```diff
@@ -16,6 +16,8 @@ COOKIE_SECRET=cookie\$ecret
 COOKIE_MAX_AGE=86400000
 REDIS_CONNECTION_STRING=
 
+ELASTICSEARCH_URL=
```
andrych, line 19: Deze nog toevoegen aan de ConfigMap, ik vermoed dat de build daarom faalt.

andrych (approved):

PR #35, opened 2022-02-11
feat(visits): add query option for searching + return name and email

bertyhell:
Added
* text search
* status filtering
* name and email of the requester

since we need those in the UI:
image

image

### DIFF
--- a/src/modules/visits/services/queries.gql.ts
+++ b/src/modules/visits/services/queries.gql.ts
@@ -17,8 +17,8 @@ export const INSERT_VISIT = `
 `;
 
 export const FIND_VISITS = `
-	query visit($offset: Int!, $limit: Int!) {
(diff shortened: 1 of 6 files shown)


ian-emsens-shd (approved):

src/modules/visits/services/queries.gql.ts
```diff
@@ -17,25 +17,32 @@ export const INSERT_VISIT = `
...
-			created_at
-			updated_at
-		}
-		cp_visit_aggregate {
-			aggregate {
-			  count
-			}
-		}
-	}
+query visit($query: String!, $offset: Int!, $limit: Int!, $statuses: [String!] = []) {
```
andrych, line 20: Klopt de indentation hier?

andrych (approved):

PR #50, opened 2022-02-22
Feature/arc 400

ian-emsens-shd:
Voegt een PATCH-route toe voor `visits`.


bertyhell:
Currently we use put calls everywhere, so best to keep using put instead of patch

There are already some put endpoints for the visits endpoint. This one seems more extensive. We should discuss what is best. one put call that can do everything, or one endpoint per usecase. Will be important once we have to check permissions for these calls.

image


You can view the swagger by surfing to: http://localhost:3100/docs

BavoVdg (approved):

bertyhell (approved):

PR #56, opened 2022-02-25
feat(arc-298): item detail from DB

bartnaessens:
Retrieve all Object details from DB

Based on PR #55 

CleanShot 2022-02-25 at 16 33 21

### DIFF
--- a/src/modules/media/services/media.service.spec.ts
+++ b/src/modules/media/services/media.service.spec.ts
@@ -26,6 +26,8 @@ const mockDataService = {
 	execute: jest.fn(),
 };
 
+const mockObjectId = objectIe.data.object_ie_by_pk.schema_identifier;
+
 const getMockMediaResponse = () => ({
 	hits: {
 		total: {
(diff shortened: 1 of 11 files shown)


src/modules/media/services/media.service.spec.ts
```diff
@@ -109,31 +110,34 @@ describe('MediaService', () => {
...
 
 	describe('findById', () => {
-		it('returns the es response from the main _search endpoint if no index is specified', async () => {
-			nock('http://elasticsearch/').post('/_search').reply(201, getMockMediaResponse());
-			const response = await mediaService.findById('123');
-			expect(response.hits.total.value).toBe(2);
-			expect(response.hits.hits.length).toBe(2);
+		it('returns the full object details as retrieved from the DB', async () => {
+			mockDataService.execute.mockResolvedValueOnce(objectIe);
+			const response = await mediaService.findById('8911p09j1g');
```
bertyhell, line 115: replace all hard coded values '8911p09j1g' with 
```
objectIe.data.object_ie_by_pk.schema_identifier
```

So if the mock changes, the test doesn't break.

bertyhell (approved):

PR #58, opened 2022-02-28
fix(tos): refactored tos date implementation

bartnaessens:
- refactored tos date implementation
- fixed test coverage

CleanShot 2022-02-28 at 11 00 21
CleanShot 2022-02-28 at 11 01 56

### DIFF
--- a/src/modules/tos/services/tos.service.spec.ts
+++ b/src/modules/tos/services/tos.service.spec.ts
@@ -1,6 +1,5 @@
 import { Test, TestingModule } from '@nestjs/testing';
 
-import tos from './__mocks__/tos';
 import { TosService } from './tos.service';
 
 import { DataService } from '~modules/data/services/data.service';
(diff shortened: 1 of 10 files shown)


src/modules/tos/services/tos.service.spec.ts
```diff
@@ -32,28 +31,45 @@ describe('TosService', () => {
 
 	describe('adapt', () => {
 		it('can adapt a hasura response to our tos interface', () => {
-			const adapted = tosService.adapt(tos);
+			const adapted = tosService.adapt('1997-01-01T00:00:00.000Z');
```
bertyhell, line 34: save this date as a mock constant at the beginning of the file

bertyhell (approved):

src/modules/tos/services/tos.service.spec.ts
bartnaessens: fixed

PR #72, opened 2022-03-07
feat(arc-431): Create check-notifications endpoint for cron job triggering of time based notifications

bertyhell:
image


src/modules/notifications/controllers/notifications.controller.ts
```diff
@@ -1,20 +1,45 @@
...
+	Headers,
+	Inject,
+	Param,
+	Patch,
+	Post,
+	Query,
+	Session,
+	UseGuards,
+} from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
```
bartnaessens, line 14: forwardRef, Inject and ConfigService are unused imports

src/config/config.const.ts
```diff
@@ -4,6 +4,7 @@ export const configValidationSchema = Joi.object({
 	NODE_ENV: Joi.string().valid('local', 'test', 'development', 'production').default('local'),
 	HOST: Joi.string(),
 	PORT: Joi.number().default(3000),
+	PROXY_API_KEY: Joi.string().required(),
```
bartnaessens, line 7: Also add to `.env/.env.template`

src/shared/guards/api-key.guard.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+const mockExecutionContextWrong = {
+	switchToHttp: jest.fn().mockReturnValue({
+		getRequest: jest.fn().mockReturnValue({
+			headers: { apiKey: 'WrongApiKey' },
+		}),
+	}),
+} as unknown as ExecutionContext;
+
+describe('ApiKeyGuard', () => {
+	it('Should allow access when apiKey header is set', async () => {
```
bartnaessens, line 44: this one fails on my machine

src/modules/notifications/controllers/notifications.controller.ts
```diff
@@ -54,4 +79,120 @@ export class NotificationsController {
 		);
 		return { status: `updated ${amountUpdated} notifications`, total: amountUpdated };
 	}
+
+	@UseGuards(ApiKeyGuard)
+	@Post('check-new-notifications')
```
bartnaessens, line 84: I'd prefer the short `@Post('check-new')`

src/modules/notifications/controllers/notifications.controller.ts
```diff
@@ -54,4 +79,120 @@ export class NotificationsController {
...
+	private async sendAccessStartNotifications(): Promise<Notification[]> {
+		const visits: Visit[] =
+			await this.visitService.getApprovedAndStartedVisitsWithoutNotification();
+		const notifications: GqlCreateOrUpdateNotification[] = visits.map(
+			(visit): GqlCreateOrUpdateNotification => ({
+				title: i18n.t('Je hebt nu toegang tot de leeszaal {{name}}', {
+					name: visit.spaceName,
+				}),
+				description: i18n.t('Je toegang vervalt terug op {{endDate}}', {
+					endDate: format(new Date(visit.endAt), 'dd/MM/yyyy HH:mm'),
```
bartnaessens, line 135: Times in DB are GMT? This is converted to timezone Europe/Brussels?

src/modules/notifications/controllers/notifications.controller.ts
```diff
@@ -54,4 +79,120 @@ export class NotificationsController {
...
+				}),
+				visit_id: visit.id,
```
bartnaessens, line 137: FE prefers snake_case instead of camelCase?

src/modules/notifications/controllers/notifications.controller.ts
```diff
@@ -54,4 +79,120 @@ export class NotificationsController {
...
+	@Post('check-new-notifications')
+	public async checkNewNotifications(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		@Headers('apiKey') apiKey: string
+	): Promise<{
+		status: string;
+		notifications?: Partial<Record<NotificationType, number>>;
+		total: number;
+	}> {
+		const accessStartsNotifications = await this.sendAccessStartNotifications();
```
bartnaessens, line 93: 🤔 shouldn't we filter these based on the logged in userId? We can't show notifications for a not-logged in user, since the frontend cannot check the recipient unless logged in. 
In that case the PROXY_API_KEY kan be discarded in favor of the loggedIn guard?
Maybe I'm missing something...

bartnaessens (changes requested):

src/modules/notifications/controllers/notifications.controller.ts
bertyhell: good catch

src/modules/notifications/controllers/notifications.controller.ts
bertyhell: yes, but this is an object that is passed into the this.notificationsService.create() function
the response of that function will use adaptNotification to correctly format the response for the frontend

src/modules/notifications/controllers/notifications.controller.ts
bertyhell: This api call will be triggered by a cron job somewhere in openshift. We do this as a route for 2 reasons:
* We can test these easily by triggering them using postman
* They can be triggered once on a single pod if the proxy every has to be load balanced across multiple pods

bartnaessens: Okay. I should add a comment above this controller function that this one is cron-triggered

bertyhell: done: https://github.com/viaacode/hetarchief-proxy/pull/72/files#diff-6e6f91f8cfa498dcec22d8053887b08861fb6df06a3532bd0725d12433e04cbaR82-R89

bartnaessens (approved):

PR #74, opened 2022-03-08
feat(arc-535): add permissions for user model

bartnaessens:
return permissions in check-login
- Requires PR for Hasura: https://github.com/viaacode/hetarchief-hasura/pull/67
- Assumes the user.profile is linked to a (user)group in the DB
CleanShot 2022-03-08 at 09 26 53
CleanShot 2022-03-08 at 09 28 23

### DIFF
--- a/src/modules/users/services/queries.gql.ts
+++ b/src/modules/users/services/queries.gql.ts
@@ -6,6 +6,13 @@ export const GET_USER_BY_IDENTITY_ID = `
 			last_name
 			mail
 			accepted_tos_at
+			group {
+				permissions {
+					permission {
+						name
+					}
+				}
+			}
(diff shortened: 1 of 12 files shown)


src/modules/users/services/queries.gql.ts
```diff
@@ -6,6 +6,13 @@ export const GET_USER_BY_IDENTITY_ID = `
 			last_name
 			mail
 			accepted_tos_at
+			group {
+				permissions {
+					permission {
```
bertyhell, line 11: might be good to sort the permissions by code, just to have a predictable list for testing equalTo() purposes

bertyhell (approved):

PR #76, opened 2022-03-08
fix(arc-311): add filter on started visits

bartnaessens:
The 'started' filter filters on visits that are currently active or were active in the past --> startAt <= now()
CleanShot 2022-03-08 at 17 18 23


bartnaessens:
Question: should the use of `started` enforce the filter on status=APPROVED?

bertyhell:
> Question: should the use of `started` enforce the filter on status=APPROVED?

No i don't think so, we might need to be pretty strict about who can use which filters based on permissions/user group, but if you get full access to the endpoint, it means you are allowed to view all of its info

bertyhell:
This might be out of scope of the task, but the reverse filter doesn't seem to work: If you set started to false i would expect it to only show the visits where startAt < now()
image

src/modules/visits/services/visits.service.ts
```diff
@@ -205,6 +215,12 @@ export class VisitsService {
 			};
 		}
 
+		if (started) {
```
bertyhell, line 218: This might be out of scope of the task, but the reverse filter doesn't seem to work: If you set started to false i would expect it to only show the visits where startAt < now()

image

```suggestion
		if (i!isUndefined(started)) {
```

src/modules/visits/services/visits.service.ts
```diff
@@ -205,6 +215,12 @@ export class VisitsService {
...
+			where.start_date = {
+				_lte: new Date().toISOString(),
+			};
+		}
```
bertyhell, line 222: This might be out of scope of the task, but the reverse filter doesn't seem to work: If you set started to false i would expect it to only show the visits where startAt < now()

image

```suggestion
		if (started) {
			where.start_date = {
				_lte: new Date().toISOString(),
			};
		} else if(!_.isUndefined(started)) {
			where.start_date = {
				_gt: new Date().toISOString(),
			};
		}
```

bertyhell (approved):

bertyhell (approved):

PR #77, opened 2022-03-09
fix(arc-294): fix query by id after db migrations

bartnaessens:


src/modules/media/services/queries.gql.ts
```diff
@@ -59,6 +58,8 @@ export const GET_OBJECT_IE_BY_ID = `
 			meemoo_media_object_id
 			schema_date_created
 			schema_date_created_lower_bound
+			ebucore_object_type
+			schema_genre
```
bertyhell, line 62: you add genre here, but not in the Media interface: 
https://github.com/viaacode/hetarchief-proxy/pull/77/files#diff-5a04c61860e3da26e7a0fdb075c87eb8c1d2ec0545d719fc524b5fea8b519aa1R76

bertyhell (approved):

src/modules/media/services/queries.gql.ts
bartnaessens: It was not added, but moved. I followed the same order of the DB columns, and it was moved there.
This way it's more easy to compare when the columns change

ian-emsens-shd (approved):

PR #78, opened 2022-03-09
fix(db): fix rename notes to visit_notes

bertyhell:


bertyhell:
it is still called "notes"
i must have had a faulty seed or outdated metadata

PR #79, opened 2022-03-09
Epic/collection interactions

ian-emsens-shd:
- Expand output of GQL collections query


src/modules/collections/services/queries.gql.ts
```diff
@@ -7,6 +7,11 @@ export const FIND_COLLECTIONS_BY_USER = `
 			is_default
 			created_at
 			updated_at
+			ies {
+				intellectual_entity {
```
bertyhell, line 11: Bart DB has changed this to ie instead of intellectual_entity
image
image


```suggestion
				ie {
```

bertyhell (changes requested):

bertyhell (approved):

PR #81, opened 2022-03-09
fix(arc-300): updated media object detail with latest db model

bartnaessens:
GET media/:id now returns the full db structure with representations and files

player ticket now requires representationId to get a playable url
CleanShot 2022-03-09 at 14 48 24
CleanShot 2022-03-09 at 14 50 57

### DIFF
--- a/src/modules/media/services/media.service.ts
+++ b/src/modules/media/services/media.service.ts
@@ -1,13 +1,13 @@
 import { Injectable, Logger, NotFoundException } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
 import got, { Got } from 'got';
-import { get, trimStart } from 'lodash';
+import { get, isEmpty, trimStart } from 'lodash';
 
 import { MediaQueryDto } from '../dto/media.dto';
 import { QueryBuilder } from '../elasticsearch/queryBuilder';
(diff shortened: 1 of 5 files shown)


src/modules/media/services/media.service.ts
```diff
@@ -155,11 +189,11 @@ export class MediaService {
...
 
 	public async getPlayableUrl(id: string, referer: string): Promise<string> {
-		const embedUrl = trimStart(await this.getEmbedUrl(id), '/');
+		const embedUrl = await this.getEmbedUrl(id);
 
 		const data = {
-			app: 'OR-avo2',
-			client: '', // TODO Required? -- SEE ARC-536
+			app: 'OR-*',
+			client: '', // Asked in ARC-536
```
bertyhell, line 196: best to leave the "TODO" part of the comment intact, so we can search for all TODO in the code

bertyhell (approved):

src/modules/media/services/media.service.ts
bertyhell, line 196:
```suggestion
			client: '', // TODO: Wait for reply on ARC-536 and implement resolution
```

bertyhell (approved):

PR #89, opened 2022-03-11
Epic/my history

ian-emsens-shd:
Adjust proxy to support https://github.com/viaacode/hetarchief-client/pull/181

### DIFF
--- a/src/modules/media/services/media.service.ts
+++ b/src/modules/media/services/media.service.ts
@@ -156,7 +156,7 @@ export class MediaService {
 				resolveBodyOnly: true,
 			});
 		} catch (e) {
-			if (e.response.statusCode === 404) {
+			if (e.response && e.response.statusCode === 404) {
 				this.logger.error(e.response.body);
 				throw new NotFoundException();
 			}
--- a/src/modules/media/services/media.service.ts
+++ b/src/modules/media/services/media.service.ts
@@ -45,8 +45,8 @@ export class MediaService {
 
 	public adapt(graphQlObject: any): Media {
 		return {
-			id: get(graphQlObject, 'meemoo_fragment_id'),
-			schemaIdentifier: get(graphQlObject, 'schema_identifier'),
+			id: get(graphQlObject, 'schema_identifier'),
+			meemooFragmentId: get(graphQlObject, 'meemoo_fragment_id'),
 			premisIdentifier: get(graphQlObject, 'premis_identifier'),
 			premisRelationship: get(graphQlObject, 'premis_relationship'),
 			isPartOf: get(graphQlObject, 'schema_is_part_of'),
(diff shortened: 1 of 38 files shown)


src/modules/visits/services/visits.service.ts
```diff
@@ -63,11 +63,21 @@ export class VisitsService {
...
+
+		return `${street}, ${postalCode} ${locality}`;
+	}
+
 	public adapt(graphQlVisit: any): Visit {
 		return {
 			id: get(graphQlVisit, 'id'),
 			spaceId: get(graphQlVisit, 'cp_space_id'),
 			spaceName: get(graphQlVisit, 'space.schema_maintainer.schema_name'),
+			spaceAddress: VisitsService.adaptSpaceAddress(graphQlVisit),
```
bertyhell, line 80: alternative so that the adaptSpaceAddress does not need to know where the address is stored in the parent object. It only needs to adapt the address that it receives.
```suggestion
	public static adaptSpaceAddress(graphQlSpaceAddress: any): string {
		const locality = get(graphQlSpaceAddress, 'locality');
		const postalCode = get(graphQlSpaceAddress, 'postal_code');
		const street = get(graphQlSpaceAddress, 'street');

		return `${street}, ${postalCode} ${locality}`;
	}

	public adapt(graphQlVisit: any): Visit {
		return {
			id: get(graphQlVisit, 'id'),
			spaceId: get(graphQlVisit, 'cp_space_id'),
			spaceName: get(graphQlVisit, 'space.schema_maintainer.schema_name'),
			spaceAddress: VisitsService.adaptSpaceAddress(get(graphQlVisit, 'space.schema_maintainer.information[0].primary_site.address')),
```

src/modules/visits/services/visits.service.ts
```diff
@@ -249,7 +259,7 @@ export class VisitsService {
 			limit,
 			orderBy: set(
 				{},
-				ORDER_PROP_TO_DB_PROP[orderProp || 'startAt'],
+				ORDER_PROP_TO_DB_PROP[orderProp] || ORDER_PROP_TO_DB_PROP['startAt'],
```
bertyhell, line 262: How is this better than what it was before? It's more code and does the exact same thing without making it much clearer.

bertyhell (approved):

src/modules/media/services/media.service.ts
```diff
@@ -157,11 +157,11 @@ export class MediaService {
 				resolveBodyOnly: true,
 			});
 		} catch (e) {
-			if (e.response.statusCode === 404) {
+			if (e.response && e.response.statusCode === 404) {
 				this.logger.error(e.response.body);
 				throw new NotFoundException();
 			}
-			this.logger.error(e.response.body);
+			this.logger.error(e.response?.body);
```
bartnaessens, line 164: the branch caused by the '?' is not hit by the test coverage

src/modules/visits/services/visits.service.ts
ian-emsens-shd: `startAt` !== `start_date`

`startAt` is not a valid value for `orderBy`

PR #92, opened 2022-03-15
feat(arc-294): refactor to a general searchfilter structure

bartnaessens:
- different free text search implementation for advanced section
- general structure: between filters === AND, inside a filter with multiValue (array): OR

CleanShot 2022-03-15 at 11 27 05

### DIFF
--- a/src/modules/media/dto/media.dto.spec.ts
+++ b/src/modules/media/dto/media.dto.spec.ts
@@ -1,10 +1,10 @@
-import { MediaQueryDto, PlayerTicketsQueryDto, SearchFilters } from '~modules/media/dto/media.dto';
+import { MediaQueryDto, PlayerTicketsQueryDto, SearchFilter } from '~modules/media/dto/media.dto';
 
 describe('NotificationsDto', () => {
 	describe('SearchFilters', () => {
 		it('should be able to construct a SearchFilters object', async () => {
-			const searchFilters = new SearchFilters();
-			expect(searchFilters).toEqual({});
+			const searchFilter = new SearchFilter();
+			expect(searchFilter).toEqual({});
 		});
 	});
 	describe('MediaQueryDto', () => {
--- a/src/modules/media/dto/media.dto.spec.ts
+++ b/src/modules/media/dto/media.dto.spec.ts
@@ -1,6 +1,6 @@
 import { MediaQueryDto, PlayerTicketsQueryDto, SearchFilter } from '~modules/media/dto/media.dto';
 
-describe('NotificationsDto', () => {
+describe('MediaDto', () => {
 	describe('SearchFilters', () => {
 		it('should be able to construct a SearchFilters object', async () => {
 			const searchFilter = new SearchFilter();
--- a/src/modules/media/dto/media.dto.ts
+++ b/src/modules/media/dto/media.dto.ts
@@ -19,17 +19,26 @@ export class SearchFilter {
 	field: SearchFilterField;
 
 	@IsArray()
+	@IsOptional()
 	@Transform((input) => {
 		if (!isArray(input.value)) {
 			return [input.value.trim()];
 		}
 		return input.value.map((kw) => kw.trim());
 	})
-	@ApiProperty({
+	@ApiPropertyOptional({
 		type: String,
-		description: `The value for the filter. Can be an array when querying for multiple values (OR operator)`,
+		description: `The array of values for the filter. Uses the OR operator. If both multiValue and value are set, value is ignored.`,
 	})
-	value: Array<string>;
+	multiValue?: Array<string>;
+
+	@IsString()
+	@IsOptional()
+	@ApiPropertyOptional({
+		type: String,
+		description: `The single value for the filter. If both multiValue and value are set, value is ignored.`,
+	})
+	value?: string;
 
 	@IsString()
 	@IsEnum(Operator)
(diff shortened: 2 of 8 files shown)


src/modules/media/dto/media.dto.spec.ts
```diff
@@ -1,10 +1,10 @@
-import { MediaQueryDto, PlayerTicketsQueryDto, SearchFilters } from '~modules/media/dto/media.dto';
+import { MediaQueryDto, PlayerTicketsQueryDto, SearchFilter } from '~modules/media/dto/media.dto';
 
 describe('NotificationsDto', () => {
```
bertyhell, line 3: should be media dto
maybe this got changes by a bad replace all somewhere?

src/modules/media/dto/media.dto.ts
```diff
@@ -1,224 +1,63 @@
...
 export class MediaQueryDto {
-	@Type(() => SearchFilters)
-	@IsOptional()
+	@Type(() => SearchFilter)
+	@IsArray()
 	@ValidateNested()
 	@ApiPropertyOptional({
-		type: SearchFilters,
-		description: 'Filters to query the media items',
+		type: SearchFilter,
```
bertyhell, line 57:
```suggestion
		type: () => [SearchFilter],
```

So swagger shows it as an array of objects:
image

src/modules/media/types.ts
```diff
@@ -120,3 +99,25 @@ export interface Media {
 	ebucoreObjectType: string;
 	representations: Representation[];
 }
+
+export enum SearchFilterField {
+	QUERY = 'query',
+	ADVANCED_QUERY = 'advancedQuery',
```
bertyhell, line 105: Weren't we going to make the client output only advanced filters, so the proxy would not need to distinguish between advanced and regular filters?
Or is this the difference between
* the search input field at the top of the page 
* and the advanced filter called "alles"?

src/modules/media/elasticsearch/consts.ts
```diff
@@ -23,32 +24,35 @@ export const READABLE_TO_ELASTIC_FILTER_NAMES: { [prop in keyof SearchFilters]:
 
 export enum QueryType {
 	TERM = 'term',
+	TERMS = 'terms',
 	RANGE = 'range',
 	MATCH = 'match',
```
bertyhell, line 29: Are these interpretations correct? If not: add correct comments
```suggestion
	TERM = 'term', // Search for a single term exactly
	TERMS = 'terms', // Must match at least one term exactly
	RANGE = 'range', // Date range or duration range
	MATCH = 'match', // Text based fuzzy search
```

src/modules/media/dto/media.dto.ts
```diff
@@ -1,224 +1,63 @@
...
-	@IsOptional()
 	@ApiPropertyOptional({
 		type: String,
-		description:
-			'The text to search for in the title, description and abstract of the media items.',
-		default: '',
+		description: `The array of values for the filter. Uses the OR operator. If both multiValue and value are set, value is ignored.`,
 	})
-	query?: string;
+	multiValue?: Array<string>;
```
bertyhell, line 33: Nice to have: have a single value param, that can either be an array or a single value. But i'm not sure if this will work:
```
@ApiProperty({
   oneOf: [
      { type: 'string' },
      { 
         type: 'array',
         items: {
            type: 'string'
         }
      }
   ]
})
```

src/modules/media/elasticsearch/queryBuilder.ts
```diff
@@ -80,116 +85,112 @@ export class QueryBuilder {
...
+				[this.getQueryType(searchFilter, value)]: {
+					[elasticKey + this.suffix(searchFilter.field)]: value,
 				},
 			},
-		}));
+		};
+	}
 
-		return result;
+	protected static buildFreeTextFilter(searchTemplate: any, searchFilter: SearchFilter): any {
```
bertyhell, line 122:
```suggestion
	protected static buildFreeTextFilter(searchTemplate: any[], searchFilter: SearchFilter): any {
```

src/modules/media/elasticsearch/queryBuilder.ts
```diff
@@ -80,116 +85,112 @@ export class QueryBuilder {
...
-		_.forEach(filters, (value: any, readableKey: keyof SearchFilters) => {
-			if (readableKey === 'query') {
-				return; // Query filter has already been handled, skip this foreach iteration
+		_.forEach(filters, (searchFilter: SearchFilter) => {
+			if (
+				searchFilter.field === SearchFilterField.QUERY ||
+				searchFilter.field === SearchFilterField.ADVANCED_QUERY
+			) {
+				if (!searchFilter.value) {
+					// empty value, ignore
```
bertyhell, line 155: wouldn't it be better to throw an error here? Just to be strict and not hide any bugs in the client?

src/modules/media/types.ts
```diff
@@ -120,3 +99,25 @@ export interface Media {
...
+	NAME = 'name',
+}
+
+export enum Operator {
+	CONTAINS = 'contains',
+	CONTAINS_NOT = 'containsNot',
+	IS = 'is',
+	IS_NOT = 'isNot',
+	GTE = 'gte',
+	LTE = 'lte',
```
bertyhell, line 122: Question: So the `between 2 dates` filter will be 2 entries in the filter array, one with operator GTE and one with LTE? ok, that is pretty clean, then we don't need to handle the `if one is set and the other isn't logic`

bertyhell (approved):

src/modules/media/types.ts
bartnaessens: Correct: 
Query = search in put field at the top of the page
Advanced query = 'alles' filter

src/modules/media/types.ts
bartnaessens: Correct!

PR #93, opened 2022-03-15
feat(arc-635): get thumbnail url

bartnaessens:
In afwachting van antwoord in https://meemoo.atlassian.net/browse/ARC-635


src/modules/media/controllers/media.controller.ts
```diff
@@ -26,6 +26,15 @@ export class MediaController {
 		return url;
 	}
 
+	@Get('thumbnail-ticket')
+	public async getThumbnailUrl(
+		@Headers('referer') referer: string,
+		@Query() thumbnailQuery: ThumbnailQueryDto
```
bertyhell, line 32: Why don't you use a route param for the representationId?
```suggestion
	@Get('thumbnail-ticket/:representationId')
	public async getThumbnailUrl(
		@Headers('referer') referer: string,
		@Param() representationId: string
```

src/modules/media/services/media.service.ts
```diff
@@ -211,14 +215,47 @@ export class MediaService {
...
+	public async getThumbnailUrl(id: string, referer: string): Promise<string> {
+		const thumbnailPath = trimStart(await this.getThumbnailPath(id), '/');
+
+		const data = {
+			app: 'OR-*',
+			client: '', // TODO: Wait for reply on ARC-536 and implement resolution
+			referer: referer || this.host,
+			maxage: this.ticketServiceMaxAge,
+		};
+		this.logger.log(data);
```
bertyhell, line 227: add todo to remove this once it runs smoothly on prod or make it a .info() call instead of a .log()

bertyhell (approved):

src/modules/media/services/media.service.spec.ts
```diff
@@ -78,10 +92,18 @@ describe('MediaService', () => {
 		expect(mediaService).toBeDefined();
 	});
 
+	describe('adaptESResponse', () => {
+		it('returns the input if no hits were found', async () => {
+			const esResponse = { hist: { total: { value: 0 } } };
```
ian-emsens-shd, line 97:
```suggestion
			const esResponse = { hits: { total: { value: 0 } } };
```

ian-emsens-shd (approved):

src/modules/media/services/tickets.service.ts
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	public async getPlayerToken(embedUrl: string, referer: string): Promise<string> {
+		// no caching
+		const token = await this.getToken(embedUrl, referer);
+		return token.jwt;
+	}
+
+	public async getThumbnailToken(referer: string): Promise<string> {
+		const thumbnailPath = '*/keyframes_all';
```
bertyhell, line 61: Discussed this with Bart DB and we should use this here:
```
testbeeld/keyframes_all
```

bertyhell (approved):

src/modules/media/services/tickets.service.ts
bartnaessens: updated

PR #98, opened 2022-03-18
feat(arc-333): Implement CM emails and trigger on changes

bartnaessens:
Send mails using CM
CleanShot 2022-03-18 at 11 43 21
CleanShot 2022-03-18 at 11 42 44
CleanShot 2022-03-18 at 11 41 16


src/modules/campaign-monitor/services/campaign-monitor.service.ts
```diff
@@ -33,12 +55,41 @@ export class CampaignMonitorService {
...
+			start_time: visit.startAt ? formatAsBelgianDate(visit.startAt, 'HH:mm') : '',
+			end_date: visit.endAt ? formatAsBelgianDate(visit.endAt, 'd MMMM yyyy') : '',
+			end_time: visit.endAt ? formatAsBelgianDate(visit.endAt, 'HH:mm') : '',
+		};
+	}
+
+	public async send(emailInfo: VisitEmailInfo): Promise<boolean> {
+		if (!emailInfo.to) {
+			// Throw exception will break too much
+			this.logger.error('Empty email address - Mail not sent', emailInfo);
```
bertyhell, line 85: this log will be useless since we won't know for which user this occured. Maybe we can include the user id in this object? or will that cause issues with CM?

bertyhell:
Should we add a test route to send an email using swagger? with apikey check. So we can verify sending mails works correctly? And maybe return the response from the CM call, then we can include this in the end-to-end tests.

Or would that be too much effort?

bertyhell:
It looks like the end var `CAMPAIGN_MONITOR_API_KEY` is still missing in the TST environment in openshift. Possibly also on the other environments.

bertyhell:
Requires change in database: https://github.com/viaacode/hetarchief-hasura/pull/85

src/modules/notifications/services/notifications.service.ts
```diff
@@ -144,32 +149,46 @@ export class NotificationsService {
...
+						name: user.firstName + ' ' + user.lastName,
+					}),
+					visit_id: visit.id,
+					type: NotificationType.NEW_VISIT_REQUEST,
+					status: NotificationStatus.UNREAD,
+				},
+				recipientIds
+			),
+			this.campaignMonitorService.send({
+				to: visit.spaceMail,
```
bertyhell, line 174: this should be the email addresses of the recipientIds that are passed to the function.
But maybe we should fetch those in the controller and pass objects instead:
```
    recipients: {id: string, email: string}[],
```

bertyhell (approved):

bartnaessens:
All keys are stored under the secrets, not in configMaps
CleanShot 2022-03-21 at 11 24 59
:

src/modules/notifications/services/notifications.service.ts
bartnaessens: updated

src/modules/campaign-monitor/services/campaign-monitor.service.ts
bartnaessens: updated

src/modules/notifications/services/notifications.service.ts
```diff
@@ -185,20 +207,29 @@ export class NotificationsService {
 					visit_id: visit.id,
 					type: NotificationType.VISIT_REQUEST_APPROVED,
 					status: NotificationStatus.UNREAD,
-					recipient: user.id,
+					recipient: user.id, // TODO question this should be the visit.visitorId, because user is the logged in user??
```
bartnaessens, line 210: @bertyhell can you check?

PR #112, opened 2022-03-28
feat(arc-531): add GET translations/nl.json

bartnaessens:
- (public) translations module: `GET translations/nl.json`. Uses caching.


src/modules/translations/services/translations.service.ts
```diff
@@ -0,0 +1,31 @@
...
+		@Inject(CACHE_MANAGER) private cacheManager: Cache
+	) {}
+
+	public async getTranslations(): Promise<Translations> {
+		let translations: Translations = await this.cacheManager.get('translations');
+		if (!translations) {
+			const response = await this.dataService.execute(GET_SITE_VARIABLES_BY_NAME, {
+				name: `translations-frontend`,
+			});
+			translations = get(response, 'data.cms_site_variables[0]');
```
bertyhell, line 25: lets throw a 404 not found if the value is undefined with message: "No translations have been set in the database"

src/modules/translations/services/translations.service.ts
```diff
@@ -0,0 +1,31 @@
...
+	constructor(
+		private dataService: DataService,
+		@Inject(CACHE_MANAGER) private cacheManager: Cache
+	) {}
+
+	public async getTranslations(): Promise<Translations> {
+		let translations: Translations = await this.cacheManager.get('translations');
+		if (!translations) {
+			const response = await this.dataService.execute(GET_SITE_VARIABLES_BY_NAME, {
+				name: `translations-frontend`,
```
bertyhell, line 23: Keep the same naming convention as with the TOS:
```suggestion
				name: `TRANSLATIONS_FRONTEND`,
```

image

src/modules/translations/services/translations.service.ts
bertyhell: Wouldn't we want to return only the values?
```suggestion
			translations = get(response, 'data.cms_site_variables[0].value');
```

bartnaessens: updated

src/modules/translations/services/translations.service.ts
bartnaessens: updated

src/modules/translations/services/queries.gql.ts
```diff
@@ -0,0 +1,8 @@
+export const GET_SITE_VARIABLES_BY_NAME = `
+	query getSiteVariableByName($name: String!) {
+		cms_site_variables(where: {name: {_eq: $name}}) {
+			name
```
bertyhell, line 4: no need to fetch name here

src/modules/translations/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface Translations {
+	name: string;
+	value: Record<string, string>;
+}
```
bertyhell, line 4:
```suggestion
export type Translations = Record<string, string>;
```

bertyhell (changes requested):

bertyhell (approved):

PR #114, opened 2022-03-29
Feature/arc 663

ian-emsens-shd:


bertyhell:
Run the test:cov command and fix the coverage for this file:
-------------------------------------------|---------|----------|---------|---------|-------------------
File                                       | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-------------------------------------------|---------|----------|---------|---------|-------------------
  spaces.service.ts                        |    90.9 |       96 |   88.88 |   90.24 | 161-170

ian-emsens-shd:
> Run the test:cov command and fix the coverage for this file:
Done, ♻️

src/modules/spaces/queries/find-space-by-cp-admin-id.graphql
```diff
@@ -0,0 +1,41 @@
+query findSpaceByCpAdminId($cpAdminId: uuid!) {
```
bertyhell, line 1: This isn't correct
A maintainer will have its own slug as a separate column. the schema_identifier is the organisation id, we used this id as a temporary placeholder
To fix this you'll have to add a slug to the space table and filter on that value instead.

src/modules/spaces/controllers/spaces.controller.ts
```diff
@@ -45,14 +45,23 @@ export class SpacesController {
 		return spaces;
 	}
 
+	@Get(':slug')
```
bertyhell, line 48: How does nestjs you know the difference between these 2 requests?
```
http://localhost:3100/spaces/my-slug
http://localhost:3100/spaces/8a8af735-a3f0-4b7f-9b9d-429e0d8fc561
```

I don't think this will be possible. But we no longer need the /spaces/:id endpoint, since everything will go though the slug. So you can remove the /:id endpoint and remove its tests

ian-emsens-shd: Removed endpoint, adjusted tests. Method on service is used in update endpoint so that still exists.

ian-emsens-shd:
♻️

bertyhell (approved):

PR #119, opened 2022-03-31
feat(arc-299): query related objects

bartnaessens:
GET /media/:schemaIdentifier/related/:meemooIdentifier

Returns all object with `meemooIdentifier`, excluding the one with current `schemaIdentifier`


src/modules/media/queries/get-related-objects.graphql
```diff
@@ -0,0 +1,77 @@
+query getRelatedObjects($schemaIdentifier: String!, $meemooIdentifier: String!) {
+	object_ie(
+		where: {
+			schema_identifier: { _neq: $schemaIdentifier }
+			meemoo_identifier: { _eq: $meemooIdentifier }
```
bertyhell, line 5: should also check that the items are inside the same maintainer:
image

bertyhell (changes requested):

src/modules/media/queries/get-related-objects.graphql
bartnaessens: fixed

src/modules/media/controllers/media.controller.ts
```diff
@@ -35,6 +35,14 @@ export class MediaController {
 		return this.mediaService.findBySchemaIdentifier(id);
 	}
 
+	@Get(':id/related/:meemooIdentifier')
```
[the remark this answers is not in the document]
bartnaessens: updated

PR #123, opened 2022-04-01
feat(thumbs): resolve thumbnail urls 

bartnaessens:
- Collections Module: for objects from collections
- Media Module: for similar and related calls


src/modules/auth/controllers/meemoo.controller.ts
```diff
@@ -113,11 +113,14 @@ export class MeemooController {
...
-					user_profile_id: archiefUser.id,
-					name: i18n.t('modules/collections/controllers___default-collection-name'),
-				});
+				await this.collectionsService.create(
+					{
+						is_default: true,
+						user_profile_id: archiefUser.id,
+						name: i18n.t('modules/collections/controllers___default-collection-name'),
+					},
+					null // referer not important here
```
ian-emsens-shd, line 122: Waarom kan `null` hier assigned worden aan een `string` functie-param?

ian-emsens-shd (approved):

src/modules/auth/controllers/meemoo.controller.ts
bartnaessens: null kan je toch altijd voor alles gebruiken? Leek me hier logischer dan '', maar is persoonlijke voorkeur wss

src/modules/collections/services/collections.service.ts
```diff
@@ -63,42 +69,56 @@ export class CollectionsService {
...
 		if (!gqlCollectionObjectLink) {
 			return undefined;
 		}
-		const func = this.adaptIeObject;
-		const objectIe = func(get(gqlCollectionObjectLink, 'ie'));
+		const objectIe = this.adaptIeObject(get(gqlCollectionObjectLink, 'ie'));
+		const resolvedThumbnailUrl = await this.playerTicketService.resolveThumbnailUrl(
+			objectIe.thumbnailUrl,
+			this.referer
+		);
```
bertyhell, line 106: This will still trigger multiple calls to the ticket service the first time when the token isn't in the cache yet.
eg:
* collection is reaquested
* collection has multiple objects
* each object tries to resolve its thumbnail in parallel
* playerTicketService.getThumbnailToken is called multiple times
* the first one that is called, will start a request to the ticket service
* network requests are slow, so the second time getThumbnailToken is called, the cache is not filled yet
* so the second call also triggers a requests, etc...

You can probably avoid this by using the .wrap() function of the cache-manager package that you are already using:
```
try {
  let token = await memoryCache.wrap(`thumbnailToken-${referer}`, function() {
    return this.getToken(thumbnailPath, referer);
  });
} catch (err) {
  // error handling
}
```

src/modules/collections/services/collections.service.ts
```diff
@@ -25,15 +25,21 @@ import {
...
+import { PlayerTicketService } from '~modules/admin/player-ticket/services/player-ticket.service';
 import { CollectionObjectsQueryDto } from '~modules/collections/dto/collections.dto';
 import { DataService } from '~modules/data/services/data.service';
 import { PaginationHelper } from '~shared/helpers/pagination';
 
 @Injectable()
 export class CollectionsService {
 	private logger: Logger = new Logger(CollectionsService.name, { timestamp: true });
+	// The current referer header
+	protected referer: string;
```
bertyhell, line 37: Why are you storing the referer inside a service? Shouldn't that be on a "per request" bases instead of across different requests?

bertyhell (changes requested):

src/modules/collections/services/collections.service.ts
bartnaessens: fixed

src/modules/collections/services/collections.service.ts
bartnaessens: fixed

src/modules/media/services/media.service.ts
```diff
@@ -21,6 +21,7 @@ import { DataService } from '~modules/data/services/data.service';
 export class MediaService {
 	private logger: Logger = new Logger(MediaService.name, { timestamp: true });
 	private gotInstance: Got;
+	private referer: string;
```
bertyhell, line 24: also a referer here, should not be saved in the service to avoid cross request contamination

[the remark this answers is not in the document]
bartnaessens: oops forgot this one indeed. Now fixed

bertyhell (approved):

bertyhell:
image

so pretty
https://tenor.com/view/proud-touched-tears-crying-man-gif-5608453

PR #130, opened 2022-04-06
Feature/arc 391

ian-emsens-shd:
https://meemoo.atlassian.net/browse/ARC-391

- expand output to include both info and description
- expand output to include slug
- add endpoint to count pending visits and fix some tests
- add error when no active visit found

Supports https://github.com/viaacode/hetarchief-client/pull/226


src/modules/visits/queries/pending-visit-count-for-user-by-slug.graphql
```diff
@@ -0,0 +1,16 @@
+query PendingVisitCountForUserBySlug($user: uuid = "", $slug: String = "") {
```
bertyhell, line 1: Avoid default values in queries since graphql won't notify us of missing variables when you do this.
It is usually better to have errors when something is missing, unless this was intended.

```suggestion
query PendingVisitCountForUserBySlug($user: uuid, $slug: String) {
```

src/modules/visits/services/visits.service.spec.ts
```diff
@@ -22,9 +22,25 @@ const getDefaultVisitsResponse = () => ({
 	},
 });
 
+const getDefaultVisitAggregateRespons = () => ({
```
bertyhell, line 25: typo
```suggestion
const getDefaultVisitAggregateResponse = () => ({
```

bertyhell (approved):

bartnaessens (approved):

PR #131, opened 2022-04-07
Feature/arc 172

ian-emsens-shd:
- Align space output with spaces service
- Expand visits service output with space info
- Adjust matching rule for maintainer id


src/modules/visits/services/queries.gql.ts
```diff
@@ -352,16 +480,35 @@ export const FIND_APPROVED_ENDED_VISITS_WITHOUT_NOTIFICATION = `
...
+				schema_image
+				schema_color
+				schema_audience_type
+				schema_description
+				schema_public_access
+				schema_service_description
+				is_published
+				published_at
+				created_at
+				updated_at
```
bertyhell, line 493: You're fetching more than you're using in the adapt function.
Is that with an eye on future features? or should we fetch less data?

bertyhell, line 493: You're fetching more than you're using in the adapt function.
Is that with an eye on future features? or should we fetch less data?

ian-emsens-shd: Consistency > needless optimisation

ian-emsens-shd: Consistency > needless optimisation

bertyhell (approved):

bartnaessens (approved):

PR #141, opened 2022-04-13
feat(arc-331): add log events

bartnaessens:
CleanShot 2022-04-13 at 15 26 45


src/modules/auth/controllers/het-archief.controller.spec.ts
```diff
@@ -165,7 +175,11 @@ describe('HetArchiefController', () => {
 			mockUsersService.getUserByIdentityId.mockReturnValueOnce(archiefUser);
 			mockIdpService.determineUserGroup.mockReturnValueOnce(Group.CP_ADMIN);
 
-			const result = await hetArchiefController.loginCallback({}, samlResponse);
+			const result = await hetArchiefController.loginCallback(
+				{ path: '/auth/hetarchief', headers: {} } as unknown as Request,
```
ian-emsens-shd, line 179: Avoid double cast, put this object into a separate const

ian-emsens-shd (approved):

src/modules/auth/controllers/het-archief.controller.spec.ts
bartnaessens: 👍

bertyhell (approved):

PR #142, opened 2022-04-14
feat(arc-296): export media object as xml

bartnaessens:
CleanShot 2022-04-14 at 14 18 36
CleanShot 2022-04-14 at 14 18 19


src/modules/media/queries/get-all-objects-by-collection-id.graphql
```diff
@@ -0,0 +1,81 @@
+query getAllObjectsByCollectionId($collectionId: uuid, $userProfileId: uuid) {
```
bertyhell, line 1: prefer the prefix "find" so it is more in line with the other queries. Also rename the file

src/modules/media/services/media.service.ts
```diff
@@ -214,6 +224,51 @@ export class MediaService {
...
 	}
 
+	/**
+	 * Get the object detail fields that are exposed as metadata
+	 */
+	public async findMetadataBySchemaIdentifier(
+		schemaIdentifier: string,
+		referer: string
+	): Promise<Partial<Media>> {
+		const object = await this.findBySchemaIdentifier(schemaIdentifier, referer);
```
bertyhell, line 234: this will resolve the thumbnail paths
to then strip them back out before exporting it to xml

maybe we can allow refer to be string | null
and if it is null, we do not resolve thumbnail urls?

Or would that be too much of an optimization?

bertyhell (approved):

src/modules/media/queries/get-all-objects-by-collection-id.graphql
bartnaessens: Renamed

PR #149, opened 2022-04-19
feat(arc-512): Move get navigation items for user outside of the admin

bertyhell:
Requires hasura PR for seed data of navigation items: https://github.com/viaacode/hetarchief-hasura/pull/114

notice: the jira item does not match with the implementation. This PR implements fetching navigation items from the database, so they can be used by the frontend to show in the header navigation bar and in the center of the footer

image


src/modules/admin/navigations/services/admin-navigations.service.ts
```diff
@@ -21,15 +21,34 @@ import {
...
 
 @Injectable()
-export class NavigationsService {
-	private logger: Logger = new Logger(NavigationsService.name, { timestamp: true });
+export class AdminNavigationsService {
+	private logger: Logger = new Logger(AdminNavigationsService.name, { timestamp: true });
 
 	constructor(private dataService: DataService) {}
 
+	public adapt(navigationItem: Navigation): NavigationItem {
```
bartnaessens, line 34: not hit by tests

src/modules/admin/navigations/services/admin-navigations.service.ts
```diff
@@ -112,7 +131,9 @@ export class NavigationsService {
 		return navigationResponse.data.app_navigation[0];
 	}
 
-	public async getNavigationElementsForUser(user: User): Promise<Record<string, Navigation[]>> {
+	public async getNavigationElementsForUser(
```
bartnaessens, line 134: not hit by tests

bartnaessens (approved):

PR #150, opened 2022-04-20
feat(arc-581): delete notifications on certain visit request updates

bartnaessens:
If startAt was changed to a future date: delete start notifications so they can be resent
Same for endAt


src/modules/notifications/dto/notifications.dto.ts
```diff
@@ -23,3 +25,18 @@ export class NotificationsQueryDto {
 	})
 	size? = 10;
 }
+
+// not yes exposed in api, used internally for consistency
```
bertyhell, line 29:
```suggestion
// not yet exposed in api, used internally for consistency
```

src/modules/notifications/services/notifications.service.ts
```diff
@@ -167,6 +170,23 @@ export class NotificationsService {
 		return affectedRows;
 	}
 
+	public async delete(visitId: string, deleteNotificationDto: DeleteNotificationDto) {
+		const response = await this.dataService.execute<DeleteNotificationsMutation>(
+			DeleteNotificationsDocument,
+			{
+				where: {
+					visit_id: { _eq: visitId },
+					type: { _in: deleteNotificationDto.types },
```
bertyhell, line 179: deleteNotificationDto.types is optional, what happens when you pass undefined?, shouldn't this default to either empty array or omit the where.type property?

bertyhell (approved):

src/modules/notifications/services/notifications.service.ts
bartnaessens: fixed

PR #151, opened 2022-04-20
feat(arc-769): Add data and path, trigger metadata export events

bertyhell:
* Add data and path options to the POST /events route
* trigger the metadata export event during export metadata for object and collection

image

### DIFF
--- a/src/modules/events/controllers/events.controller.ts
+++ b/src/modules/events/controllers/events.controller.ts
@@ -37,7 +37,6 @@ export class EventsController {
 			logEvent.data = createEventsDto.data;
 		}
 
-		// We do not await this one to not let the client wait the response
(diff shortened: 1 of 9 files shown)


src/modules/events/controllers/events.controller.ts
```diff
@@ -19,21 +20,25 @@ export class EventsController {
...
+			source: createEventsDto.path,
 			subject: user.getId(),
 			time: new Date().toISOString(),
 		};
 
+		if (createEventsDto.data) {
+			logEvent.data = createEventsDto.data;
+		}
+
 		// We do not await this one to not let the client wait the response
```
bartnaessens, line 40:
```suggestion
```

bartnaessens (approved):

PR #154, opened 2022-04-20
feat(arc-526): Create visit requests by visitor space slug instead of id

bertyhell:
So we can use the slug in the url of the client:
image


bartnaessens (approved):
Just fix 3 linting errors in visit controller/service

PR #155, opened 2022-04-21
feat(arc-443): Add permission checking for the search media endpoint

bertyhell:
* should return media objects if user is a meemoo admin
* should return media objects if user is the cp admin of the space
* should throw forbidden exception if user does not have an active visit request
* should return all media items in a specific index

image


src/modules/visits/services/visits.service.ts
```diff
@@ -406,4 +408,25 @@ export class VisitsService {
 			this.adapt(visit)
 		);
 	}
+
+	/**
+	 * Checks if the user has access to a maintainer's space because
+	 * - he is the maintainer or
+	 * - if he is a meemoo admin or
+	 * - if the visitor has an approved visit request for the current date
```
bartnaessens, line 416: Wordt nu niet enkel de laatste voorwaarde gechecked? Waar zijn de andere 2?

bertyhell: The other 2 are checked in the controller: https://github.com/viaacode/hetarchief-proxy/pull/155/files/a27f44b68750e9db70102ae1db6ffee90b54f16d#diff-94eac9f4ead4f09d561e85f8a112041f5d403eb0757f401106f887a3f33ff4c1R152-R160

I've fixed the comment to only include the one check

bartnaessens (approved):

PR #159, opened 2022-04-21
feat(arc-762): avoid directly overruled notifications

bartnaessens:
In case when a user should receive both the 'approved' and 'started' notification simultaneously, the 'approved' notification is automatically marked as read, so the users only sees the 'started' toast notification.
The approved notification is still in the list but marked as read -- or should the notification be deleted?

### DIFF
--- a/src/modules/notifications/services/notifications.service.ts
+++ b/src/modules/notifications/services/notifications.service.ts
@@ -189,6 +189,30 @@ export class NotificationsService {
 		return affectedRows;
 	}
 
+	public async simplifyUnreadNotifications(
+		notifications: Notification[],
+		userProfileId: string
+	): Promise<Notification[]> {
+		// if a user should receive both the 'approved' and 'started' notification simultaneously, only show the started one
+		const hasApproved = notifications.find(
+			(notification) =>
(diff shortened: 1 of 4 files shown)


src/modules/notifications/services/notifications.service.ts
```diff
@@ -189,6 +189,30 @@ export class NotificationsService {
...
+		const hasStarted = notifications.find(
+			(notification) =>
+				notification.type === NotificationType.ACCESS_PERIOD_READING_ROOM_STARTED &&
+				notification.status === NotificationStatus.UNREAD
+		);
+
+		if (hasApproved && hasStarted) {
+			hasApproved.status = NotificationStatus.READ;
+			await this.update(hasApproved.id, userProfileId, { status: NotificationStatus.READ });
+		}
```
bertyhell, line 211: Can't this be solved a lot easier?
Before the "approved" notification goes out, just check if the startDate has passed already, if so, set the status of the notification to "READ"

You can then trust that the "you now have access" notification will be triggered after 1min through the cron job

bertyhell:
replaced by: https://github.com/viaacode/hetarchief-proxy/pull/163

PR #167, opened 2022-04-25
feat(auth): actual user authentication logic

bartnaessens:
Update determine user logic
Removed all-permissions hack

Overview test accounts (passwords in 1password)
- your personal @studiohyperdrive.be address should be MEEMOO_ADMIN
- kiosk@vrt is the kiosk user
- support+bzt@meemoo.be is also a Meemoo Admin
- support+bztcpadmin@meemoo.be is a CP_ADMIN


bertyhell:
De account voor CP admin lijkt niet te werken op INT:
support+bztcpadmin@meemoo.be

https://idp-int.do.viaa.be/module.php/core/loginuserpass.php?

bertyhell (approved):

PR #169, opened 2022-04-26
feat(arc-509): link cp admins for archief idp

bartnaessens:
- link cp admins for archief idp
- renamed getAuth method for consistency with HetArchief controller

Issues: ARC-758 & ARC-509


src/modules/auth/controllers/het-archief.controller.ts
```diff
@@ -163,6 +164,14 @@ export class HetArchiefController {
 				}
 			}
 
+			// CP_ADMIN: Link the user to the maintainer
+			if (userGroup === Group.CP_ADMIN) {
```
bertyhell, line 168: also for kiosk users
or do you want to do that in a separate PR?

bertyhell (approved):

src/modules/auth/controllers/het-archief.controller.ts
bartnaessens: Updated in this PR

bertyhell (approved):

PR #175, opened 2022-04-28
feat(access-status): GET access-status for space and user

bartnaessens:
Returns the status should the user want to visit the space with given ID.

CleanShot 2022-04-28 at 16 56 38


src/modules/visits/types.ts
```diff
@@ -87,3 +87,9 @@ export interface Note {
 	note: string;
 	createdAt: string;
 }
+
+export interface AccessStatus {
+	spaceId: string;
+	visitorId: string;
+	status: VisitStatus;
+}
```
bertyhell, line 95: Wouldn't we just return the status, since the user known their own id and they have to provide the space id in the url.
So on the route i would return:
```
{
	status: VisitStatus;
}
```

on the service i would return:
```
VisitStatus
```

src/modules/visits/controllers/visits.controller.ts
```diff
@@ -94,9 +94,26 @@ export class VisitsController {
 		const visits = await this.visitsService.findAll(queryDto, {
 			userProfileId: user.getId(),
 		});
+
 		return visits;
 	}
 
+	@Get('space/:id/access-status')
```
bertyhell, line 101: QUESTION: should this be the id or the slug of the space? What does the client have available?

bertyhell (changes requested):
see other remarks

PR #179, opened 2022-05-03
feat(arc-670): detail call access based on object license

bartnaessens:


bartnaessens:
DRAFT: licenties voor INT en TST moeten eerst aangepast worden, anders kan geen enkel object meer bekeken worden.
Issue voor Bart DB: https://meemoo.atlassian.net/browse/ARC-800

bertyhell:
@bartnaessens Maybe we should add an environment variable: IGNORE_OBJECT_LICENSES which defaults to false.
That way we can merge this and test it out on QAS only.

bertyhell (approved):

bertyhell:
Add the env var to localhost 1password, INT and TST

This doesn't seem to work correctly locally.
I've added 
```
IGNORE_LICENSES=true
```

But on develop i get this:
image

And on this branch i get:
image

bertyhell (changes requested):
see other remark

PR #183, opened 2022-05-05
fix(access-status): use slug, bugfix query for pending requests

bartnaessens:
CleanShot 2022-05-05 at 15 56 43


src/modules/visits/services/visits.service.ts
```diff
@@ -452,11 +454,28 @@ export class VisitsService {
 				now: new Date().toISOString(),
 			}
 		);
-
-		if (!visitResponse.data.maintainer_visitor_space_request[0]) {
-			return VisitStatus.DENIED;
+		// return ACCESS, PENDING (pending or approved in the future), NO ACCESS
```
bertyhell, line 457:
```suggestion
		// return 
		// - PENDING (visit request with status pending or approved in the future)
		// - ACCESS (approved visit request with the now() time between start and end date)
		// - NO ACCESS (denied visit request or no visit request)
```

src/modules/visits/queries/find-actual-visit-by-user-and-space.graphql
```diff
@@ -3,10 +3,7 @@ query findActualVisitByUserAndSpace($userProfileId: uuid!, $spaceSlug: String!,
 		where: {
 			user_profile_id: { _eq: $userProfileId }
 			visitor_space: { slug: { _eq: $spaceSlug } }
-			_or: [
-				{ start_date: { _lte: $now }, end_date: { _gte: $now } }
-				{ status: { _eq: "PENDING" } }
-			]
+			_or: [{ end_date: { _gte: $now } }, { status: { _eq: "PENDING" } }]
```
bertyhell, line 6: This might make it easier to check in code if you only fetch pending or approved visit requests.
Then you don't have to check for canceled or denied statuses in the code.

You might also want to rename the query at that point to: findPendingOrApprovedVisitRequestsForUser
```suggestion
			_or: [{ end_date: { _gte: $now }, status: { _eq: "APPROVED" } }, { status: { _eq: "PENDING" } }]
```

bertyhell (approved):

src/modules/visits/queries/find-actual-visit-by-user-and-space.graphql
bartnaessens: 👍

PR #191, opened 2022-05-16
feat(arc-610): admin-core-permissions-check

bertyhell:
Add permission checks for graphql queries that are executed straight from the admin core client, similar to how avo queries work.


bartnaessens:
Testen falen en ik krijg die voorlopig niet gefixed

bartnaessens (approved):

PR #196, opened 2022-05-24
feat(arc-894): allow update slug

bartnaessens:
Follows https://github.com/viaacode/hetarchief-proxy/pull/194

Updates the PATCH call to allow updates for the slug


bertyhell:
swagger docs do not seem to contain the slug option:
image

PR #197, opened 2022-05-24
feat(arc-895): return status gone when space is inactive

bartnaessens:
Returns 410 Gone when no active visit, and space is inactive

CleanShot 2022-05-24 at 17 13 25

### DIFF
--- a/src/modules/visits/controllers/visits.controller.ts
+++ b/src/modules/visits/controllers/visits.controller.ts
@@ -6,6 +6,7 @@ import {
 	Controller,
 	ForbiddenException,
 	Get,
+	GoneException,
 	Logger,
 	NotFoundException,
 	Param,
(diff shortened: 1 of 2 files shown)


bertyhell:
INACTIVE spaces are still available on the homepage (GET /spaces call)

Proposal:
GET /spaces => only return active spaces
Add filter: GET /spaces?status=PENDING,ACTIVE,INACTIVE
	returns all spaces if user has permission: EDIT_ALL_SPACES
	returns ACTIVE spaces if the user does not have the permission: EDIT_ALL_SPACES

src/modules/visits/controllers/visits.controller.ts
```diff
@@ -170,6 +172,10 @@ export class VisitsController {
 			const space = await this.spacesService.findBySlug(visitorSpaceSlug);
 
 			if (space) {
+				if (space.status === VisitorSpaceStatus.Inactive) {
+					throw new GoneException();
```
bertyhell, line 176:
```suggestion
					throw new GoneException(`The space with slug '${visitorSpaceSlug}' is no longer accepting visit requests.`);
```

bertyhell (approved):

PR #210, opened 2022-06-07
fix(arc-545): correctly adapt telephone and email for a visitor space

bertyhell:
https://meemoo.atlassian.net/browse/ARC-545

image


BavoVdg (approved):
Ik close #214

PR #219, opened 2022-06-10
fix(arc-957): text search should only match with one strategy, not all

bertyhell:
https://meemoo.atlassian.net/browse/ARC-957

before: (linked to prod elasticsearch)
image

after: (linked to prod elasticsearch)
image


bartnaessens:
Fix tests :)

bartnaessens (approved):

PR #242, opened 2022-09-07
feat(avo-1083): add not found route to proxy that serves page + 404 code

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1083

The backend route http://localhost:3100/not-found returns a fake client html page, with all links going back to the client url
image


ian-emsens-shd (approved):
Why are there 2 `404.html`'s?

`src/modules/not-found/template/404.html`
`src/modules/not-found/template/modules/not-found/template/404.html`

bertyhell:
> Why are there 2 `404.html`'s?

That was an artifact from setting up the script that copied the 404.html page to the dist folder.
I removed it.

PR #246, opened 2022-09-13
Feature/arc 1095

ian-emsens-shd:
https://meemoo.atlassian.net/browse/ARC-1069

Supports https://github.com/viaacode/hetarchief-client/pull/558


src/modules/media/services/media.service.ts
```diff
@@ -414,6 +416,23 @@ export class MediaService {
...
+			{
+				meemooIdentifiers,
+			}
+		);
+
+		const count: Record<string, number> = {};
+
+		items.data?.object_ie.forEach((item) => {
+			count[item.meemoo_identifier] = count[item.meemoo_identifier]++ || 1;
+		});
```
bertyhell, line 431: Avoid the ++ operator outside of the for loop since its behavior can be confusing. In this case it has the wrong intended result. And all counts will be 1 instead of the desired count.
Since the client doesn't care, as long as the count is > 0 this isn't noticeable in the client.

Your line essentially equates to:
```
count[item.meemoo_identifier] = count[item.meemoo_identifier] = count[item.meemoo_identifier] + 1 || 1;
```

undefined + 1 => NaN => falsey

image

https://codepen.io/bertyhell/pen/zYjNrBw


```suggestion
		items.data?.object_ie.forEach((item) => {
			count[item.meemoo_identifier] = count[item.meemoo_identifier] ? count[item.meemoo_identifier] + 1 : 1;
		});
```

src/modules/media/queries/get-object-identifier-tuple.graphql
```diff
@@ -0,0 +1,6 @@
+query getObjectIdentifierTuple($meemooIdentifiers: [String!]) {
+	object_ie(where: { meemoo_identifier: { _in: $meemooIdentifiers } }) {
```
bertyhell, line 2: Optional remark: technically we don't need the count, we only need to know if there are related items

so technically this would be enough:
```
@Get('related/has-related'): boolean
```

And this query could be made more performant:
```
object_ie(where: { meemoo_identifier: { _in: $meemooIdentifiers } }, limit: 1) {
```

But having this count api call might be helpful in the future when we need to show how many related items there are.
And the performance gain is probably negligible.

bertyhell (changes requested):

src/modules/media/queries/get-object-identifier-tuple.graphql
ian-emsens-shd: To make sure this isn't always true we'd need to add an exclusion `_in` which would increase the amount of data going to-and-fro and introduce a weird edge case where two related items in the same paginated result will each have a count of 0 related items.

quick & dirty;
```
query MyQuery($_in: [String!], $_in1: [String!] = "") {
  object_ie(where: {meemoo_identifier: {_in: $_in}, _and: {_not: {schema_identifier: {_in: $_in1}}}}) {
    schema_identifier
    meemoo_identifier
  }
}
```
```
{"_in": ["k649p3zf8j", "rf5k943h2c"], "_in1": ["808d1e7077a1426b84aeef3c1da2f2b28c82f4bf1c634665bab00b3ce61008e537220edcea634da7a94749e863c0fd4b", "0efa42d7723f42fa8651c39a34bc184c7db1d2b7d0e24c01b9643eb1d03017bb8952810d0ada4893aa6824df11f023e6"]}
```

bertyhell: let leave it with the count as is. Seems easier

bertyhell (approved):

bertyhell (approved):

PR #248, opened 2022-09-19
fix(arc-1102): redact thumbnail path of inaccessible objects

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1102

before:
image
image

after:
image
image


src/modules/collections/controllers/collections.controller.ts
```diff
@@ -70,13 +74,38 @@ export class CollectionsController {
...
+		);
+
+		// Redact folder objects based on, if the user has access to that space at the moment
+		folderObjects.items.forEach((folderObject) => {
+			if (showThumbnailByMaintainerId[folderObject.maintainerId]) {
+				return;
+			}
+			// Redact object
+			delete folderObject.thumbnailUrl;
+		});
```
ian-emsens-shd, line 107: This has a lot of overlap with `src/modules/media/controllers/media.controller.ts:298`'s `applyLicensesToObject`. Can we centralise this everywhere we return an item?

ian-emsens-shd (approved):

PR #255, opened 2022-10-03
Feature/arc 1109 add pid

ian-emsens-shd:
https://meemoo.atlassian.net/browse/ARC-1109?focusedCommentId=34838


src/modules/media/services/media.service.ts
```diff
@@ -377,6 +377,9 @@ export class MediaService {
 		delete copy.meemooMediaObjectId;
 		delete copy.dateCreatedLowerBound;
 
+		// https://meemoo.atlassian.net/browse/ARC-1109?focusedCommentId=34838
+		delete copy.premisIsPartOf;
```
bertyhell, line 381: seems cleaner to do both at the same position, but it's Personal Preference
```suggestion
		copy.PID = mediaObject.premisIsPartOf;
		delete copy.premisIsPartOf;
```

bertyhell (approved):

PR #258, opened 2022-10-26
feat(avo-2187): integrate admin-core-api into hetarchief proxy

bertyhell:
https://meemoo.atlassian.net/browse/AVO-2187

requires admin-core PR: https://github.com/viaacode/react-admin-core-module/pull/48


WardVerc1 (approved):

.env/.env.template
```diff
@@ -11,11 +11,9 @@ CLIENT_API_KEY=
 CORS_ENABLE_WHITELIST=false
 
 # GRAPHQL
-GRAPHQL_URL=http://localhost:9000/v1/graphql
-GRAPHQL_SECRET=myGraphQlSecret
+GRAPHQL_URL_HET_ARCHIEF=http://localhost:9000/v1/graphql
+GRAPHQL_SECRET_HET_ARCHIEF=myGraphQlSecret
```
ian-emsens-shd, line 15: Why does this need to be namespaced in the implementing application? The admin core will only ever require one set of url/secret, it'll never need both `GRAPHQL_URL_HET_ARCHIEF` and `GRAPHQL_URL_AVO`.

src/config/config.types.ts
```diff
@@ -1,19 +1,16 @@
+import { AvoOrHetArchief } from '@meemoo/admin-core-api';
```
ian-emsens-shd, line 1: Really need to move away from these naming schemes, they're not clear nor scalable when additional applications are added.

ian-emsens-shd (approved):

.env/.env.template
bertyhell: there are multiple graphql databases:
* org api
* logging api
* hetarchief
* avo

This suffix is to make a clear difference between the hetarchief database and the logging database.

src/config/config.types.ts
bertyhell: agreed, in the future we might want to go towards: applicationName: 'avo' | 'hetarchief' | '...'
for now this name is clear, but not scalable.

PR #259, opened 2022-10-27
fix(arc-542): add route to fetch object title for seo without login

bertyhell:
https://meemoo.atlassian.net/browse/ARC-542

image

### DIFF
--- a/src/modules/media/media.types.ts
+++ b/src/modules/media/media.types.ts
@@ -12,9 +12,7 @@ export enum MediaFormat {
 	AUDIO = 'audio',
 }
 
-export type SeoInfo = {
-	title: string;
-};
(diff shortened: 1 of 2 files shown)


src/modules/media/media.types.ts
```diff
@@ -12,6 +12,10 @@ export enum MediaFormat {
 	AUDIO = 'audio',
 }
 
+export type SeoInfo = {
+	title: string;
+};
```
ian-emsens-shd, line 17: Use and extend the full type

```suggestion
export type MediaSeo = Pick<Media, 'name'>
```

ian-emsens-shd (changes requested):

PR #267, opened 2023-01-17
feat: upgrade admin-core-api

silkederudder-shd:


WardVerc1 (approved):

package.json
```diff
@@ -1,128 +1,128 @@
...
+		"redis-mock": "0.56.3",
+		"sort-object-keys": "1.1.3",
+		"source-map-support": "0.5.21",
+		"supertest": "6.1.6",
+		"ts-jest": "27.0.7",
+		"ts-loader": "9.2.6",
+		"ts-node": "10.4.0",
+		"tsconfig-paths": "3.12.0",
+		"typescript": "4.5.5"
+	}
```
bertyhell, line 127: SHD guidelines prefer 2 spaces for package.json and package-lock.json files
So we might need to update the prettier config to add a special case for these files:

```
{
  "useTabs": true,
  "tabWidth": 4,
  "trailingComma": "es5",
  "singleQuote": true,
  "endOfLine": "auto",
  "printWidth": 100,
  "overrides": [
	{
	  "files": "{package,package-lock}.json",
	  "options": {
		"useTabs": false,
		"tabWidth": 2
	  }
	}
  ]
}
```

We want to have 2 spaces for package.json, because that is how the file is formatted when you do an npm install command. And otherwise we risk sometimes committing with tabs, and sometimes with spaces

bertyhell (changes requested):
see code remark

bertyhell:
test comment

tourlbr (approved):

PR #273, opened 2023-01-23
feat(collection-limited-access): use folder access to determine limit…

tourlbr:
Screenshot 2023-01-23 at 19 44 57
Screenshot 2023-01-23 at 19 51 38
Screenshot 2023-01-23 at 19 51 52


silkederudder-shd (approved):

src/modules/collections/services/collections.service.ts
```diff
@@ -86,6 +90,18 @@ export class CollectionsService {
 		};
 	}
 
+	public adaptByVisitEndAt(visits: Pick<Visit, 'endAt'>[]): string | null {
```
bertyhell, line 93: the "adapt" is usually used to give it an object and get a similar but modified object back.
I would rename this to
```suggestion
	public getLastEndAtDate(visits: Pick<Visit, 'endAt'>[]): string | null {
```

src/modules/collections/services/collections.service.ts
```diff
@@ -105,6 +121,9 @@ export class CollectionsService {
 			createdAt: gqlCollection.created_at,
 			updatedAt: gqlCollection.updated_at,
 			isDefault: gqlCollection.is_default,
+			usedForLimitedAccessUntil:
+				this.adaptByVisitEndAt(await this.visitsService.findByFolderId(gqlCollection.id)) ||
```
bertyhell, line 125: If findByFolderId throws an error, this whole function will stop and the user will not see anything.
The usedForLimitedAccessUntil is only an informative low priority thing for the user to know. So if this fails, we should still return the rest of the folder info.

So we might want to move this to before the return and wrap it in a try catch. In the catch we can log the error to the console (console.error()). That way if it fails, it will just return undefined and not crash the whole folders call.
```
usedForLimitedAccessUntil: undefined
```

We could also do the try catch inside the findByFolderId function. But i think it is better wrapped here. Since we might want to use that function elsewhere, where we do expect an error if something goes wrong.

bertyhell (approved):

src/modules/collections/services/collections.service.ts
```diff
@@ -86,6 +90,18 @@ export class CollectionsService {
 		};
 	}
 
+	public adaptByVisitEndAt(visits: Pick<Visit, 'endAt'>[]): string | null {
+		if (isEmpty(visits) || !has(visits[0], 'endAt')) {
```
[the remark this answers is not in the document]
tourlbr: Adjusted

src/modules/collections/services/collections.service.ts
```diff
@@ -86,6 +90,18 @@ export class CollectionsService {
...
+			return null;
+		}
+
+		const reducedVisit = visits.reduce((accumulator, currentValue) =>
+			accumulator.endAt > currentValue.endAt ? accumulator : currentValue
+		);
```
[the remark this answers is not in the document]
tourlbr: Adjusted

src/modules/collections/services/collections.service.ts
tourlbr: Adjusted

src/modules/collections/services/collections.service.ts
tourlbr: Adjusted

PR #275, opened 2023-01-25
feat(objects): implement objects module

tourlbr:
Note: A new module named "ObjectsModule" was made. However the linter was complaining about using "Object" as a type so I had to use "ObjectMedia".

Screenshot 2023-01-25 at 14 20 59

### DIFF
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -15,6 +15,7 @@ import { ClientCacheModule } from '~modules/client-cache/client-cache.module';
 import { CollectionsModule } from '~modules/collections';
 import { ContentPartnersModule } from '~modules/content-partners';
 import { EventsModule } from '~modules/events';
+import { IeObjectsModule } from '~modules/ie-objects';
 import { MediaModule } from '~modules/media';
 import { NotFoundModule } from '~modules/not-found/not-found.module';
 import { NotificationsModule } from '~modules/notifications';
@@ -58,6 +59,7 @@ import { SessionService } from '~shared/services/session.service';
 		NotFoundModule,
 		AdminCoreModule,
 		MaterialRequestsModule,
+		IeObjectsModule,
 	],
 	controllers: [],
 	providers: [
--- a/src/modules/ie-objects/controllers/ie-objects.controller.ts
+++ b/src/modules/ie-objects/controllers/ie-objects.controller.ts
@@ -1,0 +1,73 @@
+import { Body, Controller, ForbiddenException, Headers, Logger, Param, Post } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import { ApiParam, ApiTags } from '@nestjs/swagger';
+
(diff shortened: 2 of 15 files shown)


src/modules/objects/objects.types.ts
```diff
@@ -0,0 +1,235 @@
...
+	alternateName: string;
+	description: string;
+	schemaIdentifier: string;
+	dctermsFormat: string;
+	transcript: string;
+	dateCreated: string;
+	files: ObjectFile[];
+}
+
+export interface ObjectMedia {
```
bertyhell, line 49: Let's use IeObject if Object throws an error
then it is closer to the database naming: objects.ie
ie => Intellectual Entity

src/modules/objects/services/objects.service.ts
```diff
@@ -0,0 +1,328 @@
...
+				hit._source.schema_thumbnail_url =
+					await this.playerTicketService.resolveThumbnailUrl(
+						hit._source.schema_thumbnail_url,
+						referer
+					);
+				return hit;
+			})
+		);
+
+		return esResponse;
```
bertyhell, line 87: the other fields can also be simplified, so we don't expose elasticsearch format to the client:
See task:
```
  "aggregations": {...},
  "items": IeObject[],
  "total": number,
```

src/modules/objects/controllers/objects.controller.ts
```diff
@@ -0,0 +1,70 @@
...
+import { ObjectsService } from '../services/objects.service';
+
+import { TranslationsService } from '~modules/translations/services/translations.service';
+import { SessionUserEntity } from '~modules/users/classes/session-user';
+import { Permission } from '~modules/users/types';
+import { RequireAllPermissions } from '~shared/decorators/require-permissions.decorator';
+import { SessionUser } from '~shared/decorators/user.decorator';
+
+@ApiTags('Objects')
+@Controller('objects')
```
bertyhell, line 18: maybe also change this then to:
```suggestion
@Controller('ie-objects')
```

src/modules/objects/controllers/objects.controller.ts
```diff
@@ -0,0 +1,70 @@
...
+export class ObjectsController {
```
bertyhell, line 19:
```suggestion
export class IeObjectsController {
```

bertyhell (approved):

src/modules/objects/objects.types.ts
tourlbr: Adjusted

src/modules/objects/controllers/objects.controller.ts
tourlbr: Adjusted

src/modules/objects/controllers/objects.controller.ts
tourlbr: Adjusted

PR #276, opened 2023-01-26
Feat/arc 1311 zendesk issue endpoint

tourlbr:
Screenshot 2023-01-26 at 11 26 45


src/modules/zendesk/controllers/zendesk.controller.ts
```diff
@@ -0,0 +1,24 @@
...
+import { RequireAnyPermissions } from '~shared/decorators/require-any-permissions.decorator';
+import { LoggedInGuard } from '~shared/guards/logged-in.guard';
+
+@UseGuards(LoggedInGuard)
+@ApiTags('Zendesk')
+@Controller('zendesk')
+export class ZendeskController {
+	@Post('/support')
+	@ApiOperation({ description: 'Create ticket through Zendesk API' })
+	@RequireAnyPermissions(Permission.CREATE_TICKET_ZENDESK)
```
bertyhell, line 18: Anonimous users can also report issues, so we should not require any permissions
We might want to limit the rate of requests to this endpoint though. eg: 1 per minute
You can use [nestjs/throttler](https://github.com/nestjs/throttler)
```suggestion
```

src/modules/zendesk/dto/zendesk.dto.ts
```diff
@@ -0,0 +1,85 @@
...
+	@IsString()
+	@Type(() => String)
+	@IsOptional()
+	@ApiPropertyOptional({
+		type: String,
+	})
+	html_body?: string | undefined;
+
+	@IsBoolean()
+	@Type(() => String)
```
bertyhell, line 31:
```suggestion
	@Type(() => Boolean)
```

bertyhell (changes requested):

PR #278, opened 2023-01-27
feat(organisations): collect sector

tourlbr:


src/modules/organisations/services/organisations.service.ts
```diff
@@ -127,6 +128,7 @@ export default class OrganisationsService implements OnApplicationBootstrap {
 				logo: organization?.logo,
 				contact_point: organization.contact_point,
 				primary_site: organization.primary_site,
+				sector: organization?.sector,
```
bertyhell, line 131: Lets default to null, so we never try to insert undefined in the database
```suggestion
				sector: organization?.sector || null,
```

bertyhell (approved):

silkederudder-shd (approved):

PR #281, opened 2023-01-31
feat(visits): add accesstype to visit endpoints

tourlbr:


src/modules/visits/dto/visits.dto.ts
```diff
@@ -155,6 +173,17 @@ export class VisitsQueryDto {
...
 	visitorSpaceSlug?: string;
 
+	@IsString()
+	@IsEnum(VisitAccessType)
+	@IsOptional()
+	@ApiPropertyOptional({
+		type: String,
+		description: 'Access type of the visit',
+		enum: VisitAccessType,
+		default: VisitAccessType.FULL,
```
bertyhell, line 183: no default, since we want to get all visit requests if no filter value was provided for accessType
```suggestion
		default: undefined,
```

src/modules/visits/services/visits.service.ts
```diff
@@ -250,11 +257,40 @@ export class VisitsService {
...
+			if (!isEmpty(accessFolderIds) && updateVisit.access_type === VisitAccessType.FOLDERS) {
+				for (let index = 0; index < accessFolderIds.length; index++) {
+					await this.dataService.execute<
+						InsertVisitFolderAccessMutation,
+						InsertVisitFolderAccessMutationVariables
+					>(InsertVisitFolderAccessDocument, {
+						folderId: accessFolderIds[index],
+						visitRequestId: currentVisit.id,
+					});
+				}
```
bertyhell, line 285: If 10 folders are selected, this code will do 10 round trips to the database before completing, which is quite slow.

You can modify the query to allow to insert multiple entries at once:
```graphql
mutation insertVisitFolderAccess($objects: [maintainer_visitor_space_request_folder_access_insert_input!] = {}) {
  insert_maintainer_visitor_space_request_folder_access(objects: $objects) {
    returning {
      id
    }
  }
}
```

There are 3 options:
* current solution: (slow)
	* make first call to db
	* receive response
	* make second call to db
	* receive response
	* make third call to db
	* receive response
	* make fourth call to db
	* receive response
* option with promise.all (fast but a lot of load on the db)
	* make first, second, third, fourth call to db
	* receive responses
* option with modified query: (fast and not a lot of load on the db)
	* make one call to the db with item 1, 2, 3, 4
	* receive response ids for 1, 2, 3, 4

src/modules/visits/types.ts
```diff
@@ -29,6 +29,11 @@ export enum VisitTimeframe {
 	FUTURE = 'FUTURE',
 }
 
+export enum VisitAccessType {
+	FULL = 'FULL',
+	FOLDERS = 'FOLDERS',
+}
```
bertyhell, line 35: You can export a renamed version of the generated database enum. That way they always stay in sync with the database enum lookup table:
```suggestion
export { Lookup_Maintainer_Visitor_Space_Request_Access_Type_Enum as VisitAccessType } from '~generated/graphql-db-types-hetarchief';
```

src/modules/visits/types.ts
```diff
@@ -76,6 +82,7 @@ export interface Visit {
 	spaceServiceDescription?: string;
 	startAt: string;
 	status: VisitStatus;
+	accessType: string;
```
bertyhell, line 85: Should we use the enum type here everywhere instead of strings?
```suggestion
	accessType: VisitAccessType;
```

bertyhell (changes requested):

src/modules/visits/queries/insert-visit-folder-access.graphql
```diff
@@ -0,0 +1,9 @@
+mutation insertVisitFolderAccess(
```
bertyhell, line 1: Optional:
```suggestion
mutation insertVisitRequestFolderAccessLinks(
```

src/modules/visits/services/__mocks__/cp_visit.ts
```diff
@@ -15,6 +16,7 @@ export const mockCpVisit: GqlVisit = {
 	user_reason: 'voor mijn onderzoek en studie',
 	user_timeframe: 'meteen',
 	status: 'PENDING',
+	access_type: 'FULL' as Lookup_Maintainer_Visitor_Space_Request_Access_Type_Enum,
```
bertyhell, line 19:
```suggestion
	access_type: VisitAccessType.FULL,
```

src/modules/visits/types.ts
```diff
@@ -53,6 +56,7 @@ export interface GqlUpdateVisit {
 	end_date: string;
 	status: VisitStatus;
 	updated_by?: string;
+	access_type?: string;
 }
```
bertyhell, line 60: always prefer the direct database type since it will avoid property naming mistakes like:
updated_by vs last_updated_by

```
export type GqlUpdateVisit = UpdateVisitMutation['update_maintainer_visitor_space_request_by_pk'];
```

src/modules/visits/dto/visits.dto.ts
```diff
@@ -92,6 +92,24 @@ export class UpdateVisitDto extends PartialType<UpdateVisitStatusDto>(UpdateVisi
...
+
+	@IsString()
+	@IsEnum(VisitAccessType)
+	@ApiPropertyOptional({
+		type: String,
+		description: 'Which access type does the visit hold access to',
+		default: VisitAccessType.Full,
+		enum: VisitAccessType,
+	})
+	accessType?: string = undefined;
```
bertyhell, line 104:
```suggestion
	accessType?: VisitAccessType = undefined;
```

bertyhell (changes requested):

PR #283, opened 2023-02-03
feat(ie-endpoints): general search endpoint

tourlbr:
Screenshot 2023-02-03 at 11 31 29

Screenshot 2023-02-03 at 11 31 36

### DIFF
--- a/src/modules/ie-objects/controllers/ie-objects.controller.ts
+++ b/src/modules/ie-objects/controllers/ie-objects.controller.ts
@@ -5,6 +5,7 @@ import { ApiParam, ApiTags } from '@nestjs/swagger';
 import { Configuration } from '~config';
 
 import { IeObjectsQueryDto } from '../dto/ie-objects.dto';
+import { checkAndFixFormatFilter } from '../helpers/check-and-fix-format-filter';
 import { IeObjectsWithAggregations } from '../ie-objects.types';
 import { IeObjectsService } from '../services/ie-objects.service';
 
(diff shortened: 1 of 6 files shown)


src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -152,6 +152,7 @@ export class IeObjectsService {
 
 	public async executeQuery(esIndex: string, esQuery: any): Promise<any> {
 		try {
+			this.logger.debug(JSON.stringify(esQuery));
```
silkederudder-shd, line 155: Remove debugger.

src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -70,4 +71,39 @@ export class IeObjectsController {
 			userHasAccessToSpace
 		);
 	}
+
+	@Post()
+	@RequireAllPermissions(Permission.SEARCH_OBJECTS)
```
bertyhell, line 76: anonymous users will also be allowed to search, so we cannot have a permission check here
```suggestion
```

src/modules/ie-objects/helpers/check-and-fix-format-filter.ts
```diff
@@ -0,0 +1,21 @@
...
+	const formatFilter = find(queryDto.filters, { field: 'format' }) as SearchFilter;
+	if (formatFilter && formatFilter.value === MediaFormat.VIDEO) {
+		// change to multivalue with video and film
+		formatFilter.multiValue = ['video', 'film'];
+		delete formatFilter.value;
+	} else if (
+		// multiValue case
+		formatFilter &&
+		formatFilter.multiValue &&
+		formatFilter.multiValue.includes(MediaFormat.VIDEO)
```
bertyhell, line 16:
```suggestion
		formatFilter?.multiValue?.includes(MediaFormat.VIDEO)
```

bertyhell (changes requested):

PR #291, opened 2023-02-15
feat(arc-1364): add new column and adjust code to work with sector props

tourlbr:
- Implement haorg_organization_type and adapt to sector
- Move retrieval of user sector from ie objects to session user

### DIFF
--- a/src/modules/auth/controllers/het-archief.controller.ts
+++ b/src/modules/auth/controllers/het-archief.controller.ts
@@ -15,7 +15,7 @@ import {
 import { ConfigService } from '@nestjs/config';
 import { ApiTags } from '@nestjs/swagger';
 import { Request, Response } from 'express';
-import { get, isEqual, pick } from 'lodash';
+import { get, isEqual, isNil, pick } from 'lodash';
 import { stringifyUrl } from 'query-string';
 
 import { Configuration } from '~config';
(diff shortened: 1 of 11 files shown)


src/modules/auth/controllers/het-archief.controller.ts
```diff
@@ -29,6 +29,7 @@ import { orgNotLinkedLogoutAndRedirectToErrorPage } from '~modules/auth/org-not-
 import { CollectionsService } from '~modules/collections/services/collections.service';
 import { EventsService } from '~modules/events/services/events.service';
 import { LogEventType } from '~modules/events/types';
+import OrganisationsService from '~modules/organisations/services/organisations.service';
```
silkederudder-shd, line 32: Consistency: use the same way of importing / exporting as the other services.

silkederudder-shd (changes requested):

src/modules/organisations/services/organisations.service.ts
```diff
@@ -85,7 +85,7 @@ export default class OrganisationsService implements OnApplicationBootstrap {
 			schemaName: gqlOrganisation?.schema_name,
 			createdAt: gqlOrganisation?.created_at,
 			updatedAt: gqlOrganisation?.updated_at,
-			sector: gqlOrganisation?.sector,
+			sector: gqlOrganisation?.haorg_organization_type,
```
bertyhell, line 88: Let's also add this link between sector and haorg_organization_type in the glossary:
https://docs.google.com/spreadsheets/d/1_vGX_pBIlJeUFt7Laoj8l3iur2qw6CtdqDs-sJ1TSL0/edit#gid=0

maybe under a new section: Organisation: represents a content partner company that does not necessarily have a visitor space. These entries come from the organisations api and are re-cached every night.

Every visitor space and every content partner should have a corresponding entry in the organisations table, but the reverse is not always true.

src/modules/auth/controllers/het-archief.controller.ts
```diff
@@ -48,7 +49,8 @@ export class HetArchiefController {
 		private collectionsService: CollectionsService,
 		private configService: ConfigService<Configuration>,
 		private eventsService: EventsService,
-		private readonly translationsService: TranslationsService
+		private readonly translationsService: TranslationsService,
+		private organisationService: OrganisationsService
```
bertyhell, line 53: they can all be readonly, or none should be readonly

bertyhell (approved):

PR #307, opened 2023-02-27
feat(arc-1410): add default metadata for objects in folders

tourlbr:
Screenshot 2023-02-27 at 16 36 07
Screenshot 2023-02-27 at 16 49 15

### DIFF
--- a/src/modules/collections/controllers/collections.controller.ts
+++ b/src/modules/collections/controllers/collections.controller.ts
@@ -102,21 +102,14 @@ export class CollectionsController {
 				referer
 			);
 
-		// Limit access to the objects in the collection
 		const visitorSpaceAccessInfo =
 			await this.ieObjectsService.getVisitorSpaceAccessInfoFromUser(user);
 
+		// Limit access to the objects in the collection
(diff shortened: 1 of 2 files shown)


src/modules/collections/controllers/collections.controller.ts
```diff
@@ -71,20 +71,15 @@ export class CollectionsController {
 		// Limit access to the objects in the collections
 		const visitorSpaceAccessInfo =
 			await this.ieObjectsService.getVisitorSpaceAccessInfoFromUser(user);
+
 		collections.items.forEach((collection) => {
-			collection.objects = (collection.objects ?? [])
-				.map((object) => {
-					return limitAccessToObjectDetails(object, {
```
bertyhell, line 77: shouldn't this change also be applied to the other GET route in the collections controller?
https://github.com/viaacode/hetarchief-proxy/blob/feat/ARC-1410/src/modules/collections/controllers/collections.controller.ts#L110

bertyhell (changes requested):
see code remark

PR #308, opened 2023-02-27
feat(arc-1221): implement extra consultable filters

tourlbr:


src/modules/ie-objects/elasticsearch/queryBuilder.ts
```diff
@@ -78,6 +73,35 @@ export class QueryBuilder {
...
+						(filter: SearchFilter) =>
+							filter.field === SearchFilterField.CONSULTABLE_REMOTE
+					),
+					isConsultableMedia: searchRequest.filters.some(
+						(filter: SearchFilter) =>
+							filter.field === SearchFilterField.CONSULTABLE_MEDIA
+					),
+				};
+
+				searchRequest.filters = searchRequest.filters.filter(
```
bertyhell, line 96: add comment:
// Remove CONSULTABLE_REMOTE and CONSULTABLE_MEDIA filter entries from the filter list, since they have been handled above and should not be handled by the standard field filtering logic.

src/modules/ie-objects/elasticsearch/queryBuilder.ts
```diff
@@ -243,7 +260,18 @@ export class QueryBuilder {
...
+				...checkSchemaLicenses,
+				{
+					terms: {
+						schema_license: [IeObjectLicense.BEZOEKERTOOL_CONTENT],
+					},
+				},
+			];
+		}
+
+		if (user.isKeyUser && !isNil(user.sector) && isConsultableRemote) {
```
bertyhell, line 274: in the if above you check:
```
if (isConsultableRemote || isConsultableMedia) {
```
but in this if, you only check one of the 2, is that intentional?
```
if (user.isKeyUser && !isNil(user.sector) && isConsultableRemote) {
```

src/modules/ie-objects/elasticsearch/queryBuilder.ts
```diff
@@ -243,7 +260,18 @@ export class QueryBuilder {
...
+		if (isConsultableRemote || isConsultableMedia) {
+			checkSchemaLicenses = [
+				...checkSchemaLicenses,
+				{
+					terms: {
+						schema_license: [IeObjectLicense.BEZOEKERTOOL_CONTENT],
+					},
+				},
+			];
+		}
```
bertyhell, line 272: **isConsultableRemote** => can i view this item's essence without visiting the visitor space.
**isOnlyConsultableOnLocation** => can i view this item's essence only by visiting the visitor space and never remotely

These 2 booleans are mutually exclusive, but it seems like you implemented **isAtLeastConsultableOnLocation** instead of either **isOnlyConsultableOnLocation** or **isConsultableRemote**

Also shouldn't **isConsultableMedia** take into account more than just the BEZOEKERTOOL_CONTENT license? It should filter for all objects where the essence is retuned, which also can happen through the VIAA-INTRA_CP-CONTENT license.

bertyhell (changes requested):

src/modules/ie-objects/elasticsearch/queryBuilder.ts
tourlbr: > isOnlyConsultableOnLocation => can i view this item's essence only by visiting the visitor space and never remotely
I don't see a way to validate this one. I have an array of visitorSpaceIds but I can't validatie with a visitor space -> that only happens when the es query is executed.

src/modules/ie-objects/elasticsearch/queryBuilder.ts
```diff
@@ -243,7 +259,29 @@ export class QueryBuilder {
...
+		if (isConsultableMedia) {
+			checkSchemaLicenses = [
+				...checkSchemaLicenses,
+				{
+					terms: {
+						schema_license: [IeObjectLicense.BEZOEKERTOOL_CONTENT],
+					},
+				},
+			];
+		}
```
bertyhell, line 282: Deze filtering lijkt niet te matchen met de spec in de story
https://meemoo.atlassian.net/wiki/spaces/HA2/pages/3705602135/Filters

image

bertyhell (changes requested):

PR #309, opened 2023-02-28
fix(arc-1231): use folder ids of the maintainer and not the requester

tourlbr:


src/modules/ie-objects/helpers/get-visitor-space-access-info-from-visits.ts
```diff
@@ -11,6 +11,6 @@ export const getVisitorSpaceAccessInfoFromVisits = (
 				?.filter((activeVisit: Visit) => activeVisit.accessType === VisitAccessType.Full)
 				?.map((activeVisit: Visit) => activeVisit.spaceMaintainerId) || [],
 		objectIds:
-			activeVisits?.flatMap((activeVisit: Visit) => activeVisit.collectionsIeSchemaIds) || [],
+			activeVisits?.flatMap((activeVisit: Visit) => activeVisit.accessibleFolderIds) || [],
```
bertyhell, line 14: This puts folder ids into a variable named: objectIds
Folders !== Objects

So you probably have to also select the objects inside those folders

bertyhell (changes requested):

silkederudder-shd (approved):

bertyhell (approved):

PR #313, opened 2023-03-02
feat(arc-1402): add boolean search business logic

tourlbr:
https://meemoo.atlassian.net/browse/ARC-1402

Screenshot 2023-03-02 at 14 45 11

Screenshot 2023-03-02 at 14 45 01


src/modules/ie-objects/helpers/convert-node-to-es-query-filter-objects.ts
```diff
@@ -0,0 +1,81 @@
...
+
+import { SearchFilter } from '../dto/ie-objects.dto';
+
+export const convertNodeToEsQueryFilterObjects = (
+	node: Expression,
+	searchTemplate?: any[],
+	searchFilter?: SearchFilter
+): any => {
+	switch (node.type) {
+		case 'Compound':
```
bertyhell, line 12: isn't there also a type: SequenceExpression that you should handle?

bertyhell (changes requested):

src/modules/ie-objects/helpers/convert-node-to-es-query-filter-objects.ts
```diff
@@ -0,0 +1,90 @@
...
+			};
+
+		case 'SequenceExpression':
+			return {
+				bool: {
+					should: ((node.expressions || []) as Expression[]).map((bodyNode) =>
+						convertNodeToEsQueryFilterObjects(bodyNode)
+					),
+				},
+			};
```
bertyhell, line 28: Both should also have a minimum_should_match equal to the number of elements in the array, otherwise this isn't an AND statements but instead an OR statement

bertyhell (approved):

PR #314, opened 2023-03-02
fix(visit-folder-access): delete folder access for update and full

tourlbr:
Issue
- During patch phase it would everytime create new entries in the folder access table. However that would cause side effect of returning more folderIds than there should be

Solution
- Check if accessType is FOLDERS with folderIds or accessType FULL -> delete existing folder access 
- For accessType is FOLDERS with folderIds insert the new access folder ids.

### DIFF
--- a/src/modules/visits/services/visits.service.ts
+++ b/src/modules/visits/services/visits.service.ts
@@ -288,18 +288,17 @@ export class VisitsService {
 				);
 			}
 
-			// IF accessFolderIds is not empty and accessType is FOLDERS
-			// THEN
-			// - remove existing access folders
-			// - add new folders to the folder_access table
-			if (!isEmpty(accessFolderIds) && updateVisit.access_type === VisitAccessType.Folders) {
-				await this.dataService.execute<
-					DeleteVisitFolderAccessMutation,
-					DeleteVisitFolderAccessMutationVariables
-				>(DeleteVisitFolderAccessDocument, {
-					visitRequestId: currentVisit.id,
-				});
+			// Always delete the old access folders
+			await this.dataService.execute<
+				DeleteVisitFolderAccessMutation,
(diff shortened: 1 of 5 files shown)


silkederudder-shd (approved):

src/modules/visits/services/visits.service.ts
```diff
@@ -283,21 +289,40 @@ export class VisitsService {
...
+			// IF accessType is FULL
+			// THEN
+			// - remove existing access folders
+			if (updateVisit.access_type === VisitAccessType.Full) {
+				await this.dataService.execute<
+					DeleteVisitFolderAccessMutation,
+					DeleteVisitFolderAccessMutationVariables
+				>(DeleteVisitFolderAccessDocument, {
+					visitRequestId: currentVisit.id,
+				});
```
bertyhell, line 325: Isn't this shorter if you always remove the folders, and then only add them if access_type is folders?
```
			// Always delete the old access folders:
			await this.dataService.execute<
				DeleteVisitFolderAccessMutation,
				DeleteVisitFolderAccessMutationVariables
			>(DeleteVisitFolderAccessDocument, {
				visitRequestId: currentVisit.id,
			});

			// IF accessFolderIds is not empty and accessType is FOLDERS
			// THEN
			// - add new folders to the folder_access table
			if (!isEmpty(accessFolderIds) && updateVisit.access_type === VisitAccessType.Folders) {
				await this.dataService.execute<
					InsertVisitFolderAccessMutation,
					InsertVisitFolderAccessMutationVariables
				>(InsertVisitFolderAccessDocument, {
					objects: [
						...accessFolderIds.map((accessFolderId: string) => ({
							folder_id: accessFolderId,
							visit_request_id: currentVisit.id,
						})),
					],
				});
			}
```

bertyhell (approved):

PR #317, opened 2023-03-06
feat(arc-1333): add missing filters to search query

tourlbr:
Added several filters to the search query json based on the dark red values in the following google sheet:

https://docs.google.com/spreadsheets/d/1xAtHfkpDi4keSsBol7pw0cQAvCmg2hWRz8oxM6cP7zo/edit#gid=0


src/modules/ie-objects/elasticsearch/templates/exact/identifier-search-query.json
```diff
@@ -26,6 +26,11 @@
           "term": {
             "premis_identifier": "{{query}}"
           }
+        },
+        {
+          "term": {
+            "schema_maintainer_id": "{{query}}"
```
bertyhell, line 32: does this field exist in elasticsearch?
shouldn't this use the id in the maintainer object? 
schema_maintainer.schema_identifier
```
...
"schema_contributor": null,
"schema_maintainer": {
    "organization_type": "Cultuur",
    "schema_identifier": "OR-154dn75",
    "alt_label": "amsab-isg",
    "schema_name": "Amsab-ISG"
},
"meemoo_local_id": "umatic_107",
...
```

src/modules/ie-objects/elasticsearch/templates/fuzzy/search-query.json
```diff
@@ -44,7 +59,13 @@
         "schema_creator",
         "schema_spatial_coverage",
         "schema_temporal_coverage",
-        "dcterms_medium"
+        "dcterms_medium",
+        "dcterm_format",
+        "ebucore_object_type",
+        "schema_publisher",
```
bertyhell, line 65: I see that you only added schema_publisher once and not for all multi-match sections. Perhaps that is on purpose because that field is not mapped as full text?
This is repeated for all multi match sections for the schema_creator field, but it is (for now) also not mapped as full text.

So we best keep both fields the same way.

Bart has expressed intention to make both fields full text, so i think we best anticipate and add both fields to all multi match sections:
* "schema_creator",
* "schema_publisher",

src/modules/ie-objects/elasticsearch/templates/fuzzy/search-query.json
```diff
@@ -44,7 +59,13 @@
         "schema_creator",
         "schema_spatial_coverage",
         "schema_temporal_coverage",
-        "dcterms_medium"
+        "dcterms_medium",
+        "dcterm_format",
```
bertyhell, line 63:
```suggestion
        "dcterms_format",
```

bertyhell (changes requested):

src/modules/ie-objects/elasticsearch/templates/fuzzy/search-query.json
tourlbr: Correct that was on purpose based on the mapping found in _mapping endpoint of ES. But I will adjusted according as you mentioned.

tourlbr: "schema_creator" already exists in all multi match fields so I'll only be adding "schema_publisher"

PR #319, opened 2023-03-07
feat(arc-1437): add form url to update cache endpoint for organisations

tourlbr:
**JIRA**
https://meemoo.atlassian.net/browse/ARC-1437

**Remark**
Use branch `feat/ARC-1435` from the `hetarchief-hasura` repo to check this feature out

### DIFF
--- a/src/modules/organisations/services/organisations.service.ts
+++ b/src/modules/organisations/services/organisations.service.ts
@@ -7,7 +7,7 @@ import {
 } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
 import got from 'got';
-import { isEmpty, isNil, uniqBy } from 'lodash';
+import { isNil, uniqBy } from 'lodash';
 
 import { Configuration } from '~config';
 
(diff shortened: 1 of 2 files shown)


src/modules/organisations/services/organisations.service.ts
```diff
@@ -101,6 +101,7 @@ export class OrganisationsService implements OnApplicationBootstrap {
     id
     description
     sector
+	form_url
```
bertyhell, line 104: indentation doesn't match the other lines

src/modules/organisations/services/organisations.service.ts
```diff
@@ -167,6 +168,7 @@ export class OrganisationsService implements OnApplicationBootstrap {
 				primary_site: organization.primary_site,
 				// Remark here organization is with Z
 				haorg_organization_type: organization?.sector || null,
+				form_url: isEmpty(organization?.form_url) ? null : organization?.form_url,
```
bertyhell, line 171: We don't really need a lodash function to have a default value for empty strings:
```suggestion
				form_url: organization?.form_url || null,
```

PR #320, opened 2023-03-07
feat(arc-1437): add form url to update cache endpoint for organisations

tourlbr:
**JIRA**
https://meemoo.atlassian.net/browse/ARC-1436

**Remark**
Use branch `feat/ARC-1435` from the `hetarchief-hasura` repo to check this feature out


src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -105,8 +105,20 @@ export class IeObjectsService {
...
+							const organisation =
+								await this.organisationService.findOrganisationBySchemaIdentifier(
+									ieObject.schemaIdentifier
+								);
+
+							return {
+								...ieObject,
+								maintainerFromUrl: organisation?.formUrl || null,
+							};
+						})
```
bertyhell, line 121: This is very expensive and not requested: Only add the form url to the detail call
You can omit it on the findAll call, since we'll never use it there.

See spec: https://meemoo.atlassian.net/browse/ARC-1436

it should only requre changes in the graphql query and the adaptFromDb function:
https://github.com/viaacode/hetarchief-proxy/pull/320/files#diff-28cd8233f726559f3280c6c212b33ec69b3d219ca06e885cde228f5ab1e8f084R310

bertyhell (changes requested):

PR #322, opened 2023-03-08
Feature/arc 1434 deny already approved visit requests

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1434

### DIFF
--- a/src/modules/visits/services/visits.service.ts
+++ b/src/modules/visits/services/visits.service.ts
@@ -243,7 +243,7 @@ export class VisitsService {
 		updateVisitDto: UpdateVisitDto,
 		userProfileId: string
 	): Promise<Visit> {
-		const { startAt, endAt, accessType /*, accessFolderIds*/ } = updateVisitDto;
+		const { startAt, endAt, accessType } = updateVisitDto;
 		let { accessFolderIds } = updateVisitDto;
 		// if any of these is set, both must be set (db constraint)
 		this.validateDates(startAt, endAt);
--- a/src/modules/visits/services/visits.service.ts
+++ b/src/modules/visits/services/visits.service.ts
@@ -243,7 +243,8 @@ export class VisitsService {
 		updateVisitDto: UpdateVisitDto,
 		userProfileId: string
 	): Promise<Visit> {
-		const { startAt, endAt, accessType, accessFolderIds } = updateVisitDto;
+		const { startAt, endAt, accessType /*, accessFolderIds*/ } = updateVisitDto;
+		let { accessFolderIds } = updateVisitDto;
 		// if any of these is set, both must be set (db constraint)
 		this.validateDates(startAt, endAt);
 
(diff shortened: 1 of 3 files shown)


src/modules/visits/services/visits.service.ts
```diff
@@ -274,44 +275,55 @@ export class VisitsService {
...
-				throw new BadRequestException(
-					`The amount of accessFolderIds (${accessFolderIds.length}) does not correspond with the accessType ${updateVisit.access_type}`
-				);
+		//IF status is DENIED
+		//THEN change access_type to FULL and clear accessFolderIds
+		if (updateVisit.status === VisitStatus.DENIED) {
+			updateVisit.access_type = VisitAccessType.Full;
+			accessFolderIds = null;
+		} else {
+			if (updateVisit.access_type) {
```
tourlbr, line 284: Wouldn't it be easier to do the following instead of change the user input?

```
if (updateVisit.access_type) {
	// Always delete the old access folders
	await this.dataService.execute<
		DeleteVisitFolderAccessMutation,
		DeleteVisitFolderAccessMutationVariables
	>(DeleteVisitFolderAccessDocument, {
		visitRequestId: currentVisit.id,
	});

	if (updateVisit.status !== VisitStatus.DENIED) {
		// IF accessFolderIds is empty and accessType is FOLDERS
		// OR accessFolderIds is not empty and accessType is FULL
		// THEN throw BadRequest exception
		if (
			isEmpty(accessType) ||
			(isEmpty(accessFolderIds) &&
			updateVisit.access_type === VisitAccessType.Folders) ||
			(!isEmpty(accessFolderIds) && updateVisit.access_type === VisitAccessType.Full)
		) {
			throw new BadRequestException(
				`The amount of accessFolderIds does not correspond with the accessType ${updateVisit.access_type}`
			);
		}

	// IF accessFolderIds is not empty and accessType is FOLDERS
	// THEN add new folders to the folder_access table
	if (
		!isEmpty(accessFolderIds) &&
		updateVisit.access_type === VisitAccessType.Folders
	) {
		await this.dataService.execute<
			InsertVisitFolderAccessMutation,
			InsertVisitFolderAccessMutationVariables
		>(InsertVisitFolderAccessDocument, {
			objects: [
				...accessFolderIds.map((accessFolderId: string) => ({
					folder_id: accessFolderId,
					visit_request_id: currentVisit.id,
				})),
			],
		});
	}
    }
}
```

@bertyhell thoughts? I just don't like setting the value after the user input

src/modules/visits/services/visits.service.ts
```diff
@@ -274,44 +275,55 @@ export class VisitsService {
...
-				(!isEmpty(accessFolderIds) && updateVisit.access_type === VisitAccessType.Full)
-			) {
-				throw new BadRequestException(
-					`The amount of accessFolderIds (${accessFolderIds.length}) does not correspond with the accessType ${updateVisit.access_type}`
-				);
+		//IF status is DENIED
+		//THEN change access_type to FULL and clear accessFolderIds
+		if (updateVisit.status === VisitStatus.DENIED) {
+			updateVisit.access_type = VisitAccessType.Full;
+			accessFolderIds = null;
```
tourlbr, line 282: REQ: this should be empty array instead of null

src/modules/visits/services/visits.service.ts
```diff
@@ -274,44 +275,55 @@ export class VisitsService {
 			}
 		}
```
tourlbr, line 277: @bertyhell Emile showed me the error but my first thought its if it is access Denied with accessType Folders FE should return accessFolderIds array with strings which does not happen. I think that is an issue on Fe side.

tourlbr (changes requested):

tourlbr:
@bertyhell Emile showed me the error but my first thought its if it is access Denied with accessType Folders FE should return accessFolderIds array with strings which does not happen. I think that is an issue on Fe side.

bertyhell:
> Emile showed me the error but my first thought its if it is access Denied with accessType Folders FE should return accessFolderIds array with strings which does not happen. I think that is an issue on Fe side.

Yeah, i was also debating where to fix it. Ideally its fixed in both. But the backend has enough info to know what to do since the status us passed as DENIED, so it shouldn't inspect any folders ids. But i do agree, that we could also fix it in the frontend.

src/modules/visits/services/visits.service.ts
bertyhell: i told Emile to do it that way, because i wanted to avoid folder ids in the database that are linked to denied visit requests.
Since that could maybe cause issues on the elasticsearch filtering side, if we don't correctly filter visit requests.

So if a visit is denied, i would always set the type to full and delete the folders.

@tourlbr Do you see any advantage in keeping those folder ids around?

tourlbr: @bertyhell No, I agree those def need to be removed

PR #327, opened 2023-03-09
Feature/arc 1450 move query parameter

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1450


src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -170,12 +172,13 @@ export class IeObjectsController {
 		return convertObjectToCsv(objectMetadata);
 	}
 
-	@Get(':schemaIdentifier/related/:meemooIdentifier')
+	@Get(':schemaIdentifier/related/:meemooIdentifier/')
```
tourlbr, line 175: REQ: don't add slash at the end of this endpoint. I presume you are having an issue getting to this endpoint when others take over. This is not the solution. 

Q: Can you explain why ending slash was added here?

src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -170,12 +172,13 @@ export class IeObjectsController {
...
 	public async getRelated(
 		@Headers('referer') referer: string,
 		@Param('schemaIdentifier') schemaIdentifier: string,
 		@Param('meemooIdentifier') meemooIdentifier: string,
-		@SessionUser() user: SessionUserEntity
+		@SessionUser() user: SessionUserEntity,
+		@Query() relatedQueryDto: RelatedQueryDto
```
tourlbr, line 181: REQ:

1. Headers
2. Param
3. Query
4. other as in SessionUser

src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -184,7 +187,8 @@ export class IeObjectsController {
 		const relatedIeObjects = await this.ieObjectsService.getRelated(
 			schemaIdentifier,
 			meemooIdentifier,
-			referer
+			referer,
+			relatedQueryDto.maintainerId
```
tourlbr, line 191: REQ: just pass the dto as full object to the service and extract maintainerId inside service.

src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -221,12 +225,17 @@ export class IeObjectsController {
 	public async getSimilar(
 		@Headers('referer') referer: string,
 		@Param('id') id: string,
-		@SessionUser() user: SessionUserEntity
+		@SessionUser() user: SessionUserEntity,
+		@Query() similarQueryDto: SimilarQueryDto
```
tourlbr, line 229: REQ: same issue as mentioned above

src/modules/ie-objects/dto/ie-objects.dto.ts
```diff
@@ -142,3 +142,23 @@ export class IeObjectMeemooIdentifiersQueryDto {
 	})
 	meemooIdentifiers: string[];
 }
+
+export class SimilarQueryDto {
```
tourlbr, line 146: REQ: Rename to IeObjectSimilarQueryDto

src/modules/ie-objects/dto/ie-objects.dto.ts
```diff
@@ -142,3 +142,23 @@ export class IeObjectMeemooIdentifiersQueryDto {
...
+	@IsString()
+	@IsOptional()
+	@ApiPropertyOptional({
+		type: String,
+		description: 'Fetch similar object with the same maintainerId',
+	})
+	maintainerId?;
+}
+
+export class RelatedQueryDto {
```
tourlbr, line 156: REQ: Rename to IeObjectRelatedQueryDto

src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -138,14 +138,16 @@ export class IeObjectsService {
 	public async getRelated(
 		schemaIdentifier: string,
 		meemooIdentifier: string,
-		referer: string
+		referer: string,
+		maintainerId?: string
```
tourlbr, line 142: REQ: should be dto and then in function extract maintainerId to be used

tourlbr (changes requested):

src/modules/ie-objects/controllers/ie-objects.controller.ts
bertyhell, line 175: Now that we figured out what the which endpoint is which, lets add some documentation to the swagger file:
```
	@ApiOperation({
		description:
			'Get objects that belong to the same meemoo identifier group. For instance objects that are highlights of the same soccer match.',
	})
```

EmileVanTichelenSHD: I first tried adding the new parameter trough the url but found it it didn't work. I thought it didn't matter whether the slash is there or not. I am gonna change it.

src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -221,12 +225,17 @@ export class IeObjectsController {
 	public async getSimilar(
```
bertyhell, line 225: Now that we figured out which endpoint is which, lets add some documentation to the swagger file:
```
	@ApiOperation({
		description:
			'Get objects that cover the same subject as the passed object schema identifier. (elasticsearch: more like this)',
	})
```

src/modules/ie-objects/controllers/ie-objects.controller.ts
tourlbr: Adding it to url would mean it is a **Param** but we need a **Query** here. So yes please remove.

src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -221,12 +225,17 @@ export class IeObjectsController {
...
+		@SessionUser() user: SessionUserEntity,
+		@Query() similarQueryDto: SimilarQueryDto
 	): Promise<IPagination<Partial<IeObject>>> {
 		const visitorSpaceAccessInfo =
 			await this.ieObjectsService.getVisitorSpaceAccessInfoFromUser(user);
 
-		const similarIeObjects = await this.ieObjectsService.getSimilar(id, '_all', referer);
+		const similarIeObjects = await this.ieObjectsService.getSimilar(
+			id,
+			similarQueryDto.maintainerId ? similarQueryDto.maintainerId.toLowerCase() : '_all',
```
bertyhell, line 236: this should be the same behavior:
```suggestion
			similarQueryDto.maintainerId?.toLowerCase() ?? '_all',
```

bertyhell (changes requested):

tourlbr (approved):

tourlbr (approved):

PR #328, opened 2023-03-10
fix issues with object limiting, null values returned from ie-objects post call

bertyhell:
## While testing with an anonymous user i saw some issues:

* null values returned from the POST /ie-objects call, which should never happen:
image

* I added a test for this:
image

* anonymous user being able to see the essence, which should not happen without the VIAA-PUBLIEK-CONTENT license
image

## After fixes:
image
image

### DIFF
--- a/src/modules/ie-objects/elasticsearch/queryBuilder.ts
+++ b/src/modules/ie-objects/elasticsearch/queryBuilder.ts
@@ -66,6 +66,7 @@ export class QueryBuilder {
 	/**
 	 * Convert filters, order, aggs properties (created by the ui) to elasticsearch query object
 	 * @param searchRequest the search query parameters
+	 * @param inputInfo
 	 * @return elastic search query
 	 */
 	public static build(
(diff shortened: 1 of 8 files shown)


src/modules/ie-objects/elasticsearch/queryBuilder.ts
```diff
@@ -228,6 +230,7 @@ export class QueryBuilder {
 							},
 						},
 					],
+					minimum_should_match: 2,
```
tourlbr, line 233: REQ: fix test cases for QueryBuilder

4 test cases are failing because of these changes please fix.

Screenshot 2023-03-10 at 20 02 16
Screenshot 2023-03-10 at 20 02 36
Screenshot 2023-03-10 at 20 02 43
Screenshot 2023-03-10 at 20 02 52
Screenshot 2023-03-10 at 20 03 01

src/modules/ie-objects/helpers/limit-access-to-object-details.ts
```diff
@@ -2,34 +2,46 @@ import { intersection, isEmpty, pick, uniq } from 'lodash';
...
+	IeObjectExtraUserGroupType,
+	IeObjectLicense,
+	IeObjectSector,
+} from '../ie-objects.types';
 
 import { getAccessThrough } from './get-access-through';
 
 import { Group } from '~modules/users/types';
 
+export interface LimitAccessUserInfo {
```
tourlbr, line 22: REQ: move interface to `ie-objects.types.ts`

tourlbr (changes requested):

tourlbr (approved):

PR #333, opened 2023-03-14
feat(arc-1486): add filter to limit aggregates

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1486

Adds a filter option to limit aggregate options based on what the user enters in the search field:
image


tourlbr (changes requested):
REQ: fix failed tests

Screenshot 2023-03-14 at 12 34 29

bertyhell:
fixed tests for query builder:
image

PR #334, opened 2023-03-14
feat(arc-1479): add is key user + extra error scenario's + setup orga…

tourlbr:
**JIRA**
https://meemoo.atlassian.net/browse/ARC-1479

**Remark**
Use hasura repo specific branch to test this out


src/modules/auth/controllers/het-archief.controller.ts
```diff
@@ -134,11 +134,13 @@ export class HetArchiefController {
 			// determine user group
 			const userGroup = await this.idpService.determineUserGroup(ldapUser);
 
+			const apps = get(ldapUser, 'attributes.apps', []);
```
bertyhell, line 137: avoid using lodash get, since it has poor type checking.
```suggestion
			const apps = ldapUser?.attributes?.apps ?? [];
```

src/modules/auth/controllers/het-archief.controller.ts
```diff
@@ -134,11 +134,13 @@ export class HetArchiefController {
...
 			const userDto = {
 				firstName: ldapUser.attributes.givenName[0],
 				lastName: ldapUser.attributes.sn[0],
 				email: ldapUser.attributes.mail[0],
 				groupId: userGroup,
+				isKeyUser: apps.includes('cataloguspro'),
```
bertyhell, line 143: maybe we should introduce an enum value for these apps?
```
export enum LdapApp {
	hetarchief= 'hetarchief',
	hetarchiefBeheer= 'hetarchief-beheer',
	cataloguspro= 'cataloguspro',
}
```

src/modules/auth/services/idp.service.ts
```diff
@@ -50,10 +50,27 @@ export class IdpService {
...
 	public async determineUserGroup(ldapUser: LdapUser): Promise<Group> {
+		const organizationalStatus = get(ldapUser, 'attributes.organizationalStatus', []);
 		// permissions check
 		const apps = get(ldapUser, 'attributes.apps', []);
+
+		// 1. organizationalStatus = kiosk + apps = hetarchief-beheer → error (account misconfiguration)
+		// 2. organizationalStatus = kiosk + apps = cataloguspro → error (account misconfiguration)
+		if (
+			organizationalStatus.includes('kiosk') &&
+			(apps.includes('hetarchief-beheer') || apps.includes('cataloguspro'))
```
bertyhell, line 65: use enum

bertyhell (approved):
I'll wait for the deploy of the database changes to be able to test this PR in real life

bertyhell:
doesn't work
i login as hetarchief2.0+bztmeemooadmin@meemoo.be
this should not be a key user and should be linked to the meemoo organisation

after login i check the database:
* user is still marked as key user
* and user doesn't have an org schema id

image

bertyhell (changes requested):
see global remark

src/modules/auth/services/idp.service.ts
```diff
@@ -50,10 +50,27 @@ export class IdpService {
...
+			(apps.includes(LdapApp.HETARCHIEF_BEHEER) || apps.includes(LdapApp.CATALOGUS_PRO))
+		) {
+			throw new Error(
+				`${NO_ORG_LINKED}${this.translationsService.t(
+					'modules/auth/services/idp___account-configuratie'
+				)}`
+			);
+		}
+
 		if (apps.includes('hetarchief-beheer')) {
```
bertyhell, line 74: use enum

bertyhell (approved):

PR #335, opened 2023-03-14
feat(arc-1482): provide separate fuzzy template to process strings wi…

tourlbr:
**Jira**
https://meemoo.atlassian.net/browse/ARC-1402
https://meemoo.atlassian.net/browse/ARC-1482

Screenshot 2023-03-14 at 16 26 41
Screenshot 2023-03-14 at 16 27 06
Screenshot 2023-03-14 at 16 27 17
Screenshot 2023-03-14 at 16 31 48

### DIFF
--- a/src/modules/ie-objects/elasticsearch/queryBuilder.ts
+++ b/src/modules/ie-objects/elasticsearch/queryBuilder.ts
@@ -376,7 +376,7 @@ export class QueryBuilder {
 						textFilters = [
 							convertNodeToEsQueryFilterObjects(
 								jsep(searchFilter.value),
-								this.config.MULTI_MATCH_QUERY_MAPPING.fuzzy.booleanQuery,
+								this.config.MULTI_MATCH_QUERY_MAPPING.exact.exactQuery,
 								searchFilter
 							),
 						];


src/modules/ie-objects/elasticsearch/templates/fuzzy/boolean-search-query.json
```diff
@@ -0,0 +1,135 @@
...
+        "schema_creator_text.lemma_split_spell",
+        "schema_spatial_coverage.lemma_split_spell",
+        "schema_temporal_coverage.lemma_split_spell",
+        "ebucore_object_type.lemma_split_spell",
+        "meemoo_description_programme.lemma_split_spell",
+        "meemoo_description_cast.lemma_split_spell",
+        "schema_publisher_text.lemma_split_spell"
+      ]
+    }
+  }
```
bertyhell, line 132: these can be removed, since we're only searching exact strings and so we don't need to split anything

src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
```diff
@@ -64,6 +66,7 @@ export enum QueryType {
 
 export const MULTI_MATCH_QUERY_MAPPING = {
 	fuzzy: {
+		booleanQuery: booleanSearchQueryTemplateFuzzy,
```
bertyhell, line 69: The other template is also used for boolean search
so we better call this one exactQuery and keep the other on "fuzzy"
```suggestion
		exactQuery: searchQueryTemplateExact,
```

bertyhell (changes requested):

bertyhell (approved):

PR #337, opened 2023-03-15
feat(arc-1464): add label as schema name through cache organisation e…

tourlbr:
**Jira**
https://meemoo.atlassian.net/browse/ARC-1464

### DIFF
--- a/src/modules/auth/controllers/het-archief.controller.ts
+++ b/src/modules/auth/controllers/het-archief.controller.ts
@@ -135,15 +135,26 @@ export class HetArchiefController {
 			const userGroup = await this.idpService.determineUserGroup(ldapUser);
 
 			const apps = ldapUser?.attributes?.apps ?? [];
+			const organisationId = isEmpty(ldapUser?.attributes?.o)
+				? null
+				: ldapUser.attributes.o[0];
+			const organisation = await this.organisationService.findOrganisationBySchemaIdentifier(
+				organisationId
+			);
 			const userDto = {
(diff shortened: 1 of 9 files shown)


src/modules/auth/controllers/het-archief.controller.ts
```diff
@@ -127,23 +127,29 @@ export class HetArchiefController {
...
+				? null
+				: ldapUser.attributes.o[0];
+			const organisation = await this.organisationService.findOrganisationBySchemaIdentifier(
+				organisationId
+			);
 			let archiefUser = await this.usersService.getUserByIdentityId(
-				ldapUser.attributes.entryUUID[0]
+				ldapUser.attributes.entryUUID[0],
+				organisation
 			);
```
bertyhell, line 140: This conflates getting something from the database, and setting something that we already know.
I would separate these in separate pieces:
```
this.usersService.getUserByIdentityId(ldapUser.attributes.entryUUID[0]) // Which internally also fetches the organisation if it is available in the avo database

const organisationId = isEmpty(ldapUser?.attributes?.o)
	? null
	: ldapUser.attributes.o[0];
const organisation = await this.organisationService.findOrganisationBySchemaIdentifier(
	organisationId
);
archiefUser.organisationId = organisationId;
archiefUser.organisationName = organisation .name;
```

That way we still have a simple function that fetches the user from the database by id (Single-responsibility principle)
and we augment that user info with the new info we received from the login ldap info afterwards.

bertyhell (approved):

PR #339, opened 2023-03-16
feat(arc-1472): add meemoo group check and add user maintainer

tourlbr:
**Jira**
https://meemoo.atlassian.net/browse/ARC-1472

### DIFF
--- a/src/modules/material-requests/services/material-requests.service.ts
+++ b/src/modules/material-requests/services/material-requests.service.ts
@@ -1,3 +1,5 @@
+import { userInfo } from 'os';
+
 import { DataService } from '@meemoo/admin-core-api';
 import { BadRequestException, Injectable, Logger, NotFoundException } from '@nestjs/common';
 import { IPagination, Pagination } from '@studiohyperdrive/pagination';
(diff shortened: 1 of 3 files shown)


src/modules/material-requests/services/material-requests.service.ts
```diff
@@ -1,3 +1,5 @@
+import { userInfo } from 'os';
```
bertyhell, line 1: is this the right import 🤔

src/modules/material-requests/services/material-requests.service.ts
```diff
@@ -99,6 +102,24 @@ export class MaterialRequestsService {
...
+			} else {
+				where.object = {
+					maintainer: {
+						schema_identifier: {
+							_in: [parameters.userMaintainerId],
+						},
+					},
+				};
+			}
+		}
```
bertyhell, line 121: PP: check access stuff in the controller and just pass a maintainerIds list to the service.

eg:
in controller check if user is cp_admin
if yes => set maintainerIds to user's maintainer id
if no => keep maintainer ids of the request params

Service just receives a simple maintainerIds list

bertyhell (changes requested):

PR #341, opened 2023-03-17
feat(arc-1316): add send route

tourlbr:
**Jira**
https://meemoo.atlassian.net/browse/ARC-1316

**Remark**
UPDATED ENV VARIABLES -> merge only when ENV variables are available on TST

**General**

17 Maart 10u05

**DONE**

- in env: CAMPAIGN_MONITOR_API_ENDPOINT opsplitsen en het ene deel als prefixurl gebruiken in de constructor van campaign-monitor.service
- in material-requests.controller.ts => move code from line 154 to 160 to adapt method in material-request.service.ts

**TODO:**

- update env variable CAMPAIGN_MONITOR_TEMPLATE_MATERIAL_REQUEST_MAINTAINER to new id → meemoo is op de hoogte en brengt dit zsm in orde
- in campaign-monitor.service.ts => in de methode 'convertMaterialRequestsToEmailTemplateData' is er een deel dat moet aangepast worden naar de nieuwe template → zie punt hierboven

**TODO in SEPERATE TiCKETS**

- de sendVisit is niet meer getest, deze moet misschien ook nog gefixt worden → [ARC-1501: BE: fix send mail for visitBACKLOG](https://meemoo.atlassian.net/browse/ARC-1501) 
- testen schrijven/aanpassen → [ARC-1500: BE: Testen schrijven voor aanvraaglijst doorsturenBACKLOG](https://meemoo.atlassian.net/browse/ARC-1500)


src/modules/campaign-monitor/services/campaign-monitor.service.ts
```diff
@@ -95,17 +100,78 @@ export class CampaignMonitorService {
...
+
+		const url = `/${
+			process.env.CAMPAIGN_MONITOR_TRANSACTIONAL_SEND_MAIL_API_VERSION as string
+		}/${process.env.CAMPAIGN_MONITOR_TRANSACTIONAL_SEND_MAIL_API_ENDPOINT as string}${
+			templateIds[emailInfo.template]
+		}/send`;
+
+		const data: CampaignMonitorData = {
+			to: recipients,
+			consentToTrack: 'unchanged',
```
bertyhell, line 138: Why is this a separate send function for material requests?

Shouldn't all this code be checked for any email we want to send? I would expect this to be in the sendTransactionalMail function.

We can keep the sendForMaterialRequest function, but only for mapping info to the email template

src/modules/material-requests/services/material-requests.service.ts
```diff
@@ -241,6 +254,46 @@ export class MaterialRequestsService {
...
+		userInfo: MaterialRequestSendRequestListUserInfo
+	): Promise<void> {
+		const groupedMaterialRequests: any = groupBy(materialRequests, 'maintainerId');
+		const groupedArray = [];
+
+		Object.keys(groupedMaterialRequests).forEach((key) => {
+			groupedArray.push(groupedMaterialRequests[key]);
+		});
+
+		// Send mail to each maintainer
```
bertyhell, line 269:
```suggestion
		// Send mail to each maintainer containing only material requests for objects they are the maintainer of
```

src/modules/campaign-monitor/dto/campaign-monitor.dto.ts
```diff
@@ -70,6 +72,68 @@ export class CampaignMonitorVisitData {
...
+	@IsOptional()
+	// @ApiPropertyOptional({ type: String })
+	user_request_context?: string;
+
+	@IsString()
+	@IsOptional()
+	// @ApiPropertyOptional({ type: String })
+	user_organisation?: string;
+}
+export class RequestListItem {
```
bertyhell, line 101:
```suggestion
}

export class RequestListItem {
```

bertyhell (changes requested):
see code remarks

src/modules/campaign-monitor/services/campaign-monitor.service.ts
EmileVanTichelenSHD: The way how the recipients are checked(line 107-113) is different than how it is in sendVisit, so this should be in the sendForMaterialRequest function. I do think that everything from line 115 to 128 could be in the sendTransactionalMail function.

EmileVanTichelenSHD:
Reason for not completely removing sendForMaterialRequest method:
This checks for specific maintainers/users that have no emailadress.
``if (emailInfo.to) {
			recipients.push(emailInfo.to);
		} else {
			if (emailInfo.isToMaintainer) {
				this.logger.error(
					`Mail will not be sent to maintainer id ${emailInfo.materialRequests[0]?.maintainerId} - empty email address`
				);
			} else {
				this.logger.error(
					`Mail will not be sent to profile id ${emailInfo.materialRequests[0]?.profileId} - empty email address`
				);
			}
		}``
While this checks if there are no recipients at all: 
``if (emailInfo.data.to.length === 0) {
				this.logger.error(
					`Mail will not be sent - no recipients. emailInfo: ${JSON.stringify(emailInfo)}`
				);
				return;
			}``
Also, moving the remaining code from sendForMaterialRequest to the materialRequestService would create more duplicate code.

EmileVanTichelenSHD (approved):

src/modules/material-requests/material-requests.types.ts
```diff
@@ -30,6 +31,7 @@ export interface MaterialRequest {
 	requesterUserGroupLabel?: string;
 	requesterUserGroupDescription?: string;
 	maintainerLogo?: string;
+	contactMail?: any;
```
bertyhell, line 34: From the fetch code i would assume this type is string | null:
```typescript
			contactMail:
				(grapqhQLMaterialRequest as FindMaterialRequestsQuery['app_material_requests'][0])
					?.object?.maintainer?.information?.contact_point || null,
```

```suggestion
	contactMail?: string | null;
```

PR #344, opened 2023-03-20
fix(is-consultable-filters): adjust logic to better reflect FE call a…

tourlbr:
After call with FE team I determined that the logic provided to determine the isConsultable filters was not sufficient and incorrect. Therefore I adjusted the existing logic to better reflect what is coming from FE.

### DIFF
--- a/codegen-hetarchief.yml
+++ b/codegen-hetarchief.yml
@@ -1,6 +1,6 @@
 overwrite: true
 schema:
-  - http://localhost:9000/v1/graphql:
+  - ${GRAPHQL_URL_HET_ARCHIEF}:
       headers:
         x-hasura-admin-secret: ${GRAPHQL_SECRET_HET_ARCHIEF}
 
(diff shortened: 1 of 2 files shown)


codegen-hetarchief.yml
```diff
@@ -1,6 +1,6 @@
 overwrite: true
 schema:
-  - ${GRAPHQL_URL_HET_ARCHIEF}:
+  - http://localhost:9000/v1/graphql:
```
bertyhell, line 3: should this be committed? maybe its better to default it to localhost:9000 indeed, but then i would add the old line in a comment, so we can still pull types from the tst database if bart changes something and we don't have the hetarchief hasura running locally

bertyhell (changes requested):

codegen-hetarchief.yml
tourlbr: I was a bit to eager to add stuff to the commit message. This should indeed not be committed.

PR #346, opened 2023-03-21
Feature/arc 1500 request list tests

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1500

I also changed some things from send request list to make it cleaner / more optimized.


src/modules/campaign-monitor/services/campaign-monitor.service.spec.ts
```diff
@@ -69,8 +81,10 @@ const getMockVisit = (): Visit => ({
 
 describe('CampaignMonitorService', () => {
 	let campaignMonitorService: CampaignMonitorService;
+	const env = process.env;
```
tourlbr, line 84: REQ: this should no longer be required so please remove this.

src/modules/campaign-monitor/services/campaign-monitor.service.spec.ts
```diff
@@ -69,8 +81,10 @@ const getMockVisit = (): Visit => ({
...
 
 	beforeEach(async () => {
+		process.env = { ...env };
```
tourlbr, line 87: REQ: this should no longer be required so please remove this.

tourlbr (changes requested):

tourlbr (approved):

src/modules/campaign-monitor/templates/templates.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Template } from '../campaign-monitor.types';
+
+export class TemplateService {
```
tourlbr, line 3: REQ: You can just do similar logic as you dit with xml & csv. No need for Class.

tourlbr (changes requested):

tourlbr (approved):

bertyhell (approved):

src/modules/campaign-monitor/services/campaign-monitor.service.ts
```diff
@@ -203,15 +204,15 @@ export class CampaignMonitorService {
 
 	public async sendTransactionalMail(
 		emailInfo: CampaignMonitorSendMailDto
-	): Promise<void | BadRequestException> {
+	): Promise<void | BadRequestException | boolean> {
```
bertyhell, line 207: We shouldn't add exceptions in the return type
```suggestion
	): Promise<boolean> {
```

bertyhell (changes requested):

bertyhell (changes requested):

PR #356, opened 2023-03-23
fix(arc-1539): set is pending false for successfull material requests…

tourlbr:
**Jira**
https://meemoo.atlassian.net/browse/ARC-1539

### DIFF
--- a/src/modules/material-requests/controllers/material-requests.controller.ts
+++ b/src/modules/material-requests/controllers/material-requests.controller.ts
@@ -3,6 +3,7 @@ import {
 	Controller,
 	Delete,
 	Get,
+	InternalServerErrorException,
 	Param,
 	Patch,
 	Post,
@@ -23,6 +24,10 @@ import {
 import { MaterialRequest, MaterialRequestMaintainer } from '../material-requests.types';
 import { MaterialRequestsService } from '../services/material-requests.service';
 
+import {
+	Lookup_App_Material_Request_Requester_Capacity_Enum,
+	Lookup_App_Material_Request_Type_Enum,
+} from '~generated/graphql-db-types-hetarchief';
 import { SessionUserEntity } from '~modules/users/classes/session-user';
 import { Group, Permission } from '~modules/users/types';
 import { RequireAnyPermissions } from '~shared/decorators/require-any-permissions.decorator';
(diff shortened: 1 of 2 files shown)


src/modules/material-requests/services/material-requests.service.ts
```diff
@@ -204,16 +204,17 @@ export class MaterialRequestsService {
...
 	): Promise<MaterialRequest> {
-		const { type, reason, organisation, requester_capacity } = materialRequestDto;
+		const { type, reason, organisation, requester_capacity, is_pending } = materialRequestDto;
 
 		const updateMaterialRequest = {
 			type,
 			reason,
 			organisation,
 			requester_capacity,
+			is_pending,
```
bertyhell, line 217: We don't want to expose the is_pending to the api
This is set to false when you send your material requests, but we don't want anyone to be able to change this directly from the client side.

src/modules/material-requests/controllers/material-requests.controller.ts
```diff
@@ -154,29 +159,53 @@ export class MaterialRequestsController {
...
+				materialRequests.items.map(async (materialRequest: MaterialRequest) => {
+					await this.materialRequestsService.updateMaterialRequest(
+						materialRequest.id,
+						user.getId(),
+						{
+							type: materialRequest.type as Lookup_App_Material_Request_Type_Enum,
+							reason: materialRequest.reason,
+							organisation: materialRequest.organisation,
+							requester_capacity:
+								materialRequest.requesterCapacity as Lookup_App_Material_Request_Requester_Capacity_Enum,
```
bertyhell, line 195: Shouldn't the MaterialRequest interface already contain:
```
export interface MaterialRequest {
	id: string;
	objectSchemaIdentifier: string;
	objectSchemaName: string;
	objectMeemooIdentifier: string;
	objectMeemooLocalId?: string;
	objectDctermsFormat: MediaFormat;
	objectThumbnailUrl: string;
	profileId: string;
	reason: string;
	createdAt: string;
	updatedAt: string;
	type: string;
	isPending: boolean;
	requesterId: string;
	requesterFullName: string;
	requesterMail: string;
        requesterCapacity: Lookup_App_Material_Request_Requester_Capacity_Enum
	maintainerId: string;
	maintainerName: string;
	maintainerSlug: string;
	requesterUserGroupId?: string;
	requesterUserGroupName?: string;
	requesterUserGroupLabel?: string;
	requesterUserGroupDescription?: string;
	maintainerLogo?: string;
	contactMail?: any;
	organisation?: string | null;
}
```
So we don't need to cast this here anymore.

bertyhell (changes requested):

PR #358, opened 2023-03-27
fix(arc-1563): return cp-admin space and visits for permanent access

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1563

Needed to get this frontend issue to work correctly: https://meemoo.atlassian.net/browse/ARC-1176

CP_ADMINs should always have access to their own space.

- Fixed for /spaces route
- Fixed for the /visits/personal route
- Fixed some minor bugs around user groups

image

### DIFF
--- a/src/modules/users/services/users.service.spec.ts
+++ b/src/modules/users/services/users.service.spec.ts
@@ -1,4 +1,4 @@
-import { DataService, Group } from '@meemoo/admin-core-api';
+import { DataService } from '@meemoo/admin-core-api';
 import { Test, TestingModule } from '@nestjs/testing';
 
 import { UsersService } from './users.service';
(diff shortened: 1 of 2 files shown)


src/modules/users/services/users.service.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { DataService } from '@meemoo/admin-core-api';
+import { DataService, Group } from '@meemoo/admin-core-api';
```
tourlbr, line 1: Q: shouldn't this one be removed? 🤔

bertyhell: yesn good catch

EmileVanTichelenSHD (approved):

silkederudder-shd (approved):

PR #359, opened 2023-03-28
feat(arc-1510): add filter to search by maintainer name

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1510

### DIFF
--- a/src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
+++ b/src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
@@ -33,8 +33,9 @@ export enum SearchFilterField {
 	DESCRIPTION = 'description',
 	ERA = 'era',
 	LOCATION = 'location',
-	MAINTAINER_ID = 'maintainerId', // Contains the OR-id of the maintainer
-	MAINTAINER_NAME = 'maintainerName', // Contains the name of the maintainer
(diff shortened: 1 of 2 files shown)


silkederudder-shd (approved):

src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
```diff
@@ -33,7 +33,9 @@ export enum SearchFilterField {
 	DESCRIPTION = 'description',
 	ERA = 'era',
 	LOCATION = 'location',
-	MAINTAINER = 'maintainer',
+	// TODO future: rename maintainer to maintainerId and maintainers to maintainerName and also change this in the client
```
tourlbr, line 36: REQ: If we are adding it anyway can't we just add it already the correct way?

tourlbr (changes requested):

src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
bertyhell: no, this requires quite a lot of changes on the client, since this value is used all through the filtering mechanism

tourlbr: REQ: add ticket for this in Jira

tourlbr (approved):

PR #360, opened 2023-03-28
fix(arc-1510): do not add keyword for maintainer id filter

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1510

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "start": "nest start",
     "__predev": "npm run generate-database-types",
     "prestart:debug": "npm run generate-database-types",
-    "dev": "dotenv -e .env/.env.local nest start --watch",
+    "dev": "dotenv -e .env/.env.local nest start -- --watch",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "start": "nest start",
     "__predev": "npm run generate-database-types",
     "prestart:debug": "npm run generate-database-types",
-    "dev": "dotenv -e .env/.env.local nest start --watch",
+    "dev": "dotenv -e .env/.env.local nest start -- --watch",
```
silkederudder-shd, line 21: Do we need the extra `--` ?

silkederudder-shd (changes requested):

package.json
bertyhell: yes, i think npm doesn't otherwise correctly pass the --watch all the way to the next start commend.
Without this fix the server doesn't auto restart
with this fix the server does correctly restart

PR #363, opened 2023-03-29
Feature/arc 1454 sign in newsletter

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1454

New in env:
```
CAMPAIGN_MONITOR_TEMPLATE_CONFIRMATION=f33d92b0-f185-495c-820c-f09983b82f15
CAMPAIN_MONITOR_CONFIRM_EMAIL_TOKEN_SECRET_KEY=secretKey
CAMPAIN_MONITOR_CONFIRM_EMAIL_TOKEN_SECRET_IV=secretIV
CAMPAIN_MONITOR_CONFIRM_EMAIL_TOKEN_ECNRYPTION_METHOD=aes-256-cbc
```


src/modules/campaign-monitor/campaign-monitor.types.ts
```diff
@@ -47,3 +48,14 @@ export interface CampaignMonitorShareFolderInfo {
 export interface CampaignMonitorNewsletterPreferences {
 	newsletter: boolean;
 }
+
+export interface UserInfo {
```
tourlbr, line 52: REQ: rename to **CampaignMonitorUserInfo**

Tip: Always use module name first and then custom name. We do this because this would not interfere with any other modules that would have similar interface names

src/modules/campaign-monitor/helpers/crypto-helper.ts
```diff
@@ -0,0 +1,41 @@
+import crypto from 'crypto';
```
tourlbr, line 1: Q: are these helpers really needed 🤔 

@bertyhell thoughts on this approach?

tourlbr (changes requested):

src/modules/campaign-monitor/helpers/crypto-helper.ts
EmileVanTichelenSHD: The Jira ticket had a guide on how to do the encryption/decryption which uses crypto. Crypto was already added as a package.

tourlbr (approved):

bertyhell (changes requested):

src/modules/campaign-monitor/helpers/crypto-helper.ts
```diff
@@ -0,0 +1,41 @@
...
+	);
+	return Buffer.from(cipher.update(data, 'utf8', 'hex') + cipher.final('hex')).toString('base64'); // Encrypts data and converts to hex and base64
+};
+// Decrypt data
+export const decryptData = (encryptedData: string): string => {
+	const buff = Buffer.from(encryptedData, 'base64');
+	const decipher = crypto.createDecipheriv(
+		encryptionParams().encryption_method,
+		encryptionParams().key,
+		encryptionParams().encryptionIV
```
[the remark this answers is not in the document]
EmileVanTichelenSHD: This is resolved, the button just doesn't work.

PR #364, opened 2023-03-30
Feature/arc 1541 sync custom fields

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1541

In the env I changed CAMPAIGN_MONITOR_OPTIN_LIST_HETARCHIEF to the actual id.

### DIFF
--- a/src/modules/auth/controllers/het-archief.controller.spec.ts
+++ b/src/modules/auth/controllers/het-archief.controller.spec.ts
@@ -11,6 +11,7 @@ import { IdpService } from '../services/idp.service';
 
 import { HetArchiefController } from './het-archief.controller';
 
+import { CampaignMonitorService } from '~modules/campaign-monitor/services/campaign-monitor.service';
 import { CollectionsService } from '~modules/collections/services/collections.service';
 import { EventsService } from '~modules/events/services/events.service';
 import { OrganisationsService } from '~modules/organisations/services/organisations.service';
@@ -108,6 +109,11 @@ const mockOrganisationsService: Partial<Record<keyof OrganisationsService, jest.
 	findOrganisationBySchemaIdentifier: jest.fn(),
 };
 
+const mockCampaignMonitorService: Partial<Record<keyof CampaignMonitorService, jest.SpyInstance>> =
+	{
+		updateNewsletterPreferences: jest.fn(),
+	};
+
 const mockRequest = { path: '/auth/hetarchief', headers: {} } as unknown as Request;
 
 const getNewMockSession = () => ({
@@ -157,6 +163,10 @@ describe('HetArchiefController', () => {
 				{
 					provide: OrganisationsService,
 					useValue: mockOrganisationsService,
+				},
+				{
+					provide: CampaignMonitorService,
+					useValue: mockCampaignMonitorService,
 				},
 			],
 		})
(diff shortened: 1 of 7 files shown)


src/modules/auth/controllers/het-archief.controller.spec.ts
```diff
@@ -234,12 +244,35 @@ describe('HetArchiefController', () => {
...
+				samlResponse,
+				{}
+			);
+
+			expect(result).toEqual({
+				statusCode: HttpStatus.TEMPORARY_REDIRECT,
+				url: hetArchiefLoginUrl,
+			});
+			expect(mockUsersService.createUserWithIdp).not.toBeCalled();
+			expect(mockUsersService.updateUser).not.toBeCalled();
```
bertyhell, line 267: also add this check for the other tests:
```suggestion
			expect(mockUsersService.updateUser).not.toBeCalled();
			expect(mockCampaignMonitorService.updateNewsletterPreferences).toBeCalledTimes(1);
```

src/modules/auth/controllers/het-archief.controller.ts
```diff
@@ -226,6 +228,22 @@ export class HetArchiefController {
...
+					email: archiefUser?.email,
+					is_key_user: archiefUser?.isKeyUser,
+					usergroup: archiefUser?.groupName,
+					created_date: archiefUser?.createdAt,
+					last_access_date: archiefUser?.lastAccessAt,
+					organisation: archiefUser?.organisationName,
+				});
+			} catch (err) {
+				this.logger.error('Failed updating the custom fields to Campaign Monitor');
+			}
```
bertyhell, line 245: This is an example where we don't want to wait for campaign monitor to allow the user to login.
Its already good that you added a try catch. But if something goes wrong with campaign monitor and requests have to timeout before they fail, the login could take a long time.
So we better not wait, but still catch the error. You can do that like this:
```suggestion
			
				this.campaignMonitorService.updateNewsletterPreferences({
					firstName: archiefUser?.firstName,
					lastName: archiefUser?.lastName,
					email: archiefUser?.email,
					is_key_user: archiefUser?.isKeyUser,
					usergroup: archiefUser?.groupName,
					created_date: archiefUser?.createdAt,
					last_access_date: archiefUser?.lastAccessAt,
					organisation: archiefUser?.organisationName,
				}).catch(err => {
					this.logger.error('Failed updating the custom fields to Campaign Monitor: ' + JSON.stringify(err));
				})
```

src/modules/campaign-monitor/services/campaign-monitor.service.spec.ts
```diff
@@ -386,6 +386,42 @@ describe('CampaignMonitorService', () => {
...
+			);
+			expect(result.CustomFields[3]).toEqual(
+				mockNewsletterTemplateDataWithNewsletter.CustomFields[3]
+			);
+			expect(result.CustomFields[4]).toEqual(
+				mockNewsletterTemplateDataWithNewsletter.CustomFields[4]
+			);
+			expect(result.CustomFields[6]).toEqual(
+				mockNewsletterTemplateDataWithNewsletter.CustomFields[6]
+			);
```
bertyhell, line 422: did you forget index 5, or is it not needed to be checked?
In that case add a comment to the code. eg:
```
// field [5] is .... and we don't care about that this value is because ....
```

src/modules/campaign-monitor/services/campaign-monitor.service.spec.ts
```diff
@@ -525,12 +552,27 @@ describe('CampaignMonitorService', () => {
...
+					)}/${mockConfigService.get(
+						'CAMPAIGN_MONITOR_SUBSCRIBER_API_ENDPOINT'
+					)}/${mockConfigService.get('CAMPAIGN_MONITOR_OPTIN_LIST_HETARCHIEF')}.json`
+				)
+				.reply(201, {});
+
+			try {
+				await campaignMonitorService.updateNewsletterPreferences(mockUserInfo);
 			} catch (e) {
 				expect(e).toBeUndefined();
```
bertyhell, line 577: This should be a jest fail() statement, since we always want to fail, even if e is undefined, since it would mean that the updateNewsletterPreferences still threw an error causing it to execute the catch statement

bertyhell (approved):

src/modules/campaign-monitor/services/campaign-monitor.service.spec.ts
EmileVanTichelenSHD: We don't want this test to fail. When it fails, e is defined and the test failed.

PR #365, opened 2023-03-30
feat(arc-1552): return http 410 for inactive spaces

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1552

### DIFF
--- a/src/modules/spaces/controllers/spaces.controller.spec.ts
+++ b/src/modules/spaces/controllers/spaces.controller.spec.ts
@@ -100,19 +100,19 @@ describe('SpacesController', () => {
 		it('should return all spaces', async () => {
 			mockSpacesService.findAll.mockResolvedValueOnce(mockSpacesResponse);
 			const spaces = await spacesController.getSpaces({}, new SessionUserEntity(mockUser));
-			expect(spaces.items.length).toEqual(3);
+			expect(spaces.items.length).toEqual(mockSpacesResponse.items.length);
 		});
 
 		it('should return all spaces if no user is logged in', async () => {
 			mockSpacesService.findAll.mockResolvedValueOnce(mockSpacesResponse);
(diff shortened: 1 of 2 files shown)


src/modules/spaces/controllers/spaces.controller.spec.ts
```diff
@@ -94,19 +100,19 @@ describe('SpacesController', () => {
 		it('should return all spaces', async () => {
 			mockSpacesService.findAll.mockResolvedValueOnce(mockSpacesResponse);
 			const spaces = await spacesController.getSpaces({}, new SessionUserEntity(mockUser));
-			expect(spaces.items.length).toEqual(2);
+			expect(spaces.items.length).toEqual(3);
```
bertyhell, line 103: If we use the length of the mock array, we never have to update this again. And it explains where the "3" value comes from
```suggestion
			expect(spaces.items.length).toEqual(mockSpacesResponse.items.length);
```

src/modules/spaces/controllers/spaces.controller.spec.ts
```diff
@@ -168,6 +174,22 @@ describe('SpacesController', () => {
...
+			try {
+				await spacesController.getSpaceBySlug('huis-van-alijn');
+			} catch (err) {
+				error = err;
+			}
+			expect(error?.response).toEqual({
+				statusCode: 410,
+				message: 'Space with slug "huis-van-alijn" is inactive',
+				error: 'Gone',
+			});
```
bertyhell, line 191: You could simplify this to:
```suggestion
			try {
				await spacesController.getSpaceBySlug('huis-van-alijn');
				fail(...)
			} catch (err) {
				expect(err?.response).toEqual({
					statusCode: 410,
					message: 'Space with slug "huis-van-alijn" is inactive',
					error: 'Gone',
				});
			}
```

bertyhell (approved):

tourlbr (approved):

PR #366, opened 2023-03-30
feat(arc-1550): adjust spaces logic for admin

tourlbr:
**Jira**
https://meemoo.atlassian.net/browse/ARC-1550

**Remark**
FE should hide purple block cause is not representative for meemoo admin

Screenshot 2023-03-30 at 15 02 49
Screenshot 2023-03-30 at 15 02 54
Screenshot 2023-03-30 at 15 03 04


bertyhell:
I committed a change to revert the changes for the spaces endpoint since this was also affecting the admin screens in the admin dashboard causing the filters to stop working:
image

For spaces i'll make a PR for the frontend to change the query params to request all spaces instead of the backend forcing the meemoo admin to receive all the spaces

PR #368, opened 2023-04-03
Feature/arc 1579 organisations tests

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1579

### DIFF
--- a/src/modules/organisations/mocks/organisations.mocks.ts
+++ b/src/modules/organisations/mocks/organisations.mocks.ts
@@ -34,6 +34,8 @@ export const mockGqlOrganisation: GqlOrganisation = {
 		'De Vlaamse Radio- en Televisieomroeporganisatie, afgekort VRT, is de Nederlandstalige openbare omroep voor radio en televisie in België.',
 	// Remark here organization is with Z
 	haorg_organization_type: IeObjectSector.PUBLIC,
+	form_url:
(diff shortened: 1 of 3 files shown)


src/modules/organisations/mocks/organisations.mocks.ts
```diff
@@ -0,0 +1,106 @@
...
+			street: 'Auguste Reyerslaan 52 B,',
+			telephone: '+32 2 741 37 20',
+			post_office_box_number: '2',
+		},
+	},
+	schemaName: 'VRT',
+	createdAt: '2023-04-03T13:28:01.434203+02:00',
+	updatedAt: '2023-04-03T13:28:01.434203+02:00',
+	sector: IeObjectSector.PUBLIC,
+	formUrl: undefined,
```
tourlbr, line 65: REQ: use empty string for this prop. You should never set undefined yourself if you want to set something than use null.

tourlbr (changes requested):

src/modules/organisations/mocks/organisations.mocks.ts
bertyhell: null would be prefered here indeed

bertyhell (approved):

tourlbr (approved):

PR #370, opened 2023-04-04
Feature/arc 1533 change events triggers

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1533

### DIFF
--- a/src/modules/auth/controllers/meemoo.controller.ts
+++ b/src/modules/auth/controllers/meemoo.controller.ts
@@ -164,7 +164,8 @@ export class MeemooController {
 					time: new Date().toISOString(),
 					data: {
 						idp: Idp.MEEMOO,
-						user_group: archiefUser.groupId,
+						user_group_name: archiefUser.groupName,
+						user_group_id: archiefUser.groupId,
 					},
 				},
 			]);
--- a/src/modules/auth/controllers/meemoo.controller.ts
+++ b/src/modules/auth/controllers/meemoo.controller.ts
@@ -164,6 +164,7 @@ export class MeemooController {
 					time: new Date().toISOString(),
 					data: {
 						idp: Idp.MEEMOO,
+						user_group: archiefUser.groupId,
 					},
 				},
 			]);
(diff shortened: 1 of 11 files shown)


tourlbr (approved):

src/modules/auth/controllers/meemoo.controller.ts
```diff
@@ -164,6 +164,7 @@ export class MeemooController {
 					time: new Date().toISOString(),
 					data: {
 						idp: Idp.MEEMOO,
+						user_group: archiefUser.groupId,
```
bertyhell, line 167: it's better to use groupName instead of groupId since those uuid don't mean anything outside of the hetarchief database. So the groupName (eg: CP_ADMIN) will be more meaningful

src/modules/campaign-monitor/controllers/campaign-monitor.controller.ts
```diff
@@ -50,11 +57,27 @@ export class CampaignMonitorController {
...
+		// Log event
+		this.eventsService.insertEvents([
+			{
+				id: EventsHelper.getEventId(request),
+				type: LogEventType.NEWSLETTER_SUBSCRIBE,
+				source: request.path,
+				subject: user?.getId() || 'anonymous',
+				time: new Date().toISOString(),
+				data: {
+					user_group: user?.getGroupId() || 'anonymous',
```
bertyhell, line 73:
```suggestion
					user_group: user?.getGroupName() || 'ANONYMOUS',
```

src/modules/events/controllers/events.controller.ts
```diff
@@ -36,6 +36,11 @@ export class EventsController {
 			logEvent.data = createEventsDto.data;
 		}
 
+		logEvent.data = {
+			...logEvent.data,
+			user_group: user.getGroupId() || 'anonymous',
```
bertyhell, line 41: same remark

bertyhell (changes requested):

bertyhell (approved):

PR #374, opened 2023-04-05
fix(arc-1591): fix accessthrough for meemoo admin

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1591

User without access:
image

Meemoo admin:
image

### DIFF
--- a/src/modules/ie-objects/controllers/ie-objects.controller.ts
+++ b/src/modules/ie-objects/controllers/ie-objects.controller.ts
@@ -29,7 +29,7 @@ import { IeObjectsService } from '../services/ie-objects.service';
 import { EventsService } from '~modules/events/services/events.service';
 import { LogEventType } from '~modules/events/types';
 import { SessionUserEntity } from '~modules/users/classes/session-user';
-import { GroupId, Permission } from '~modules/users/types';
+import { GroupName, Permission } from '~modules/users/types';
 import { VisitsService } from '~modules/visits/services/visits.service';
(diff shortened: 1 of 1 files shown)


src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -86,6 +88,17 @@ export class IeObjectsController {
 			accessibleVisitorSpaceIds: visitorSpaceAccessInfo.visitorSpaceIds,
 		});
 
+		// Meemoo admin user always has VISITOR_SPACE_FULL in accessThrough when object has BEZOEKTOOL licences
+		if (
+			user.getGroupId() === GroupId.MEEMOO_ADMIN &&
```
tourlbr, line 93:
```suggestion
			user.getGroupName() === GroupName.MEEMOO_ADMIN &&
```

src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -86,6 +88,17 @@ export class IeObjectsController {
 			accessibleVisitorSpaceIds: visitorSpaceAccessInfo.visitorSpaceIds,
 		});
 
+		// Meemoo admin user always has VISITOR_SPACE_FULL in accessThrough when object has BEZOEKTOOL licences
```
tourlbr, line 91:
```suggestion
		// Meemoo admin user always has VISITOR_SPACE_FULL in accessThrough when object has BEZOEKERTOOL licences
```

tourlbr (changes requested):

PR #375, opened 2023-04-05
fix(arc-1590): mock visitorspaceRequest for meemoo admin

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1590

### DIFF
--- a/src/modules/visits/controllers/visits.controller.ts
+++ b/src/modules/visits/controllers/visits.controller.ts
@@ -226,6 +226,7 @@ export class VisitsController {
 		@SessionUser() user: SessionUserEntity
 	): Promise<Visit | null> {
 		// Check if the user is a CP admin or a Kiosk user for the requested space
+		// MEEMOO_ADMIN has access to all the visitor spaces
 		if (
 			visitorSpaceSlug === user.getVisitorSpaceSlug() ||
 			user.getGroupName() === GroupName.MEEMOO_ADMIN


src/modules/visits/controllers/visits.controller.ts
```diff
@@ -226,7 +226,10 @@ export class VisitsController {
 		@SessionUser() user: SessionUserEntity
 	): Promise<Visit | null> {
 		// Check if the user is a CP admin or a Kiosk user for the requested space
-		if (visitorSpaceSlug === user.getVisitorSpaceSlug()) {
+		if (
+			visitorSpaceSlug === user.getVisitorSpaceSlug() ||
+			user.getGroupId() === GroupId.MEEMOO_ADMIN
```
bertyhell, line 231: At some point we would like to get rid of the GroupId enum since it contains uuids and we don't like hardcoding uuids in the code. So prefer GroupName where possible.
```suggestion
			user.getGroupName() === GroupName.MEEMOO_ADMIN
```

Maybe you can also add a depracted notice to the GroupId enum, so it warns other devs to not use it anymore:
```
/**
 * @deprecated At some point we would like to get rid of the GroupId enum since it contains uuids and we don't like hardcoding uuids in the code. So prefer GroupName where possible.
 */
export enum GroupId...
```

bertyhell (changes requested):

src/modules/visits/controllers/visits.controller.ts
```diff
@@ -226,7 +226,10 @@ export class VisitsController {
 		@SessionUser() user: SessionUserEntity
 	): Promise<Visit | null> {
 		// Check if the user is a CP admin or a Kiosk user for the requested space
```
tourlbr, line 228: REQ: this comment line doesn't seem to be accurate anymore please adjust

tourlbr (changes requested):

src/modules/visits/controllers/visits.controller.ts
bertyhell: Yes it should be moved to line 262
and this comment should read:
// MEEMOO_ADMIN has access to all the visitor spaces

bertyhell (approved):

tourlbr (approved):

PR #376, opened 2023-04-05
fix(arc-1595): give cp admin access to own space and meemoo admin to all

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1595


bertyhell:
merging voor demo christophe
reviews are still welcome

PR #380, opened 2023-04-06
Bugfix/arc 1597 share folder objects

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1597


src/modules/collections/controllers/collections.controller.ts
```diff
@@ -290,6 +291,33 @@ export class CollectionsController {
...
+			// TODO: make it possible to insert all objects to the new collection at once
+			await Promise.all(
+				folderObjects?.items.map(async (item) => {
+					await this.collectionsService.addObjectToCollection(
+						createdCollection.id,
+						item?.schemaIdentifier,
+						referer
+					);
+				})
+			);
```
bertyhell, line 312: This can potentially fire off houndreds of requests to the database
Lets limit the simultanious number of events to 20 at a time:
```
import * as promiseUtils from 'blend-promise-utils';

await promiseUtils.mapLimit(folderObjects?.items, 20, async (item) => {
	await this.collectionsService.addObjectToCollection(
		createdCollection.id,
		item?.schemaIdentifier,
		referer
	);
});
```

src/modules/collections/controllers/collections.controller.ts
```diff
@@ -290,6 +291,33 @@ export class CollectionsController {
...
+		} catch (err) {
+			if (err.name !== 'NotFoundException') {
```
bertyhell, line 314: also add a console.error with the actual error, so we can debug problems on the server by looking at the logs

bertyhell (changes requested):
see code comments

src/modules/collections/controllers/collections.controller.ts
```diff
@@ -290,6 +292,32 @@ export class CollectionsController {
...
+			// TODO: make it possible to insert all objects to the new collection at once
+			await promiseUtils.mapLimit(folderObjects?.items, 20, async (item) => {
+				await this.collectionsService.addObjectToCollection(
+					createdCollection.id,
+					item?.schemaIdentifier,
+					referer
+				);
+			});
+		} catch (err) {
+			if (err.name !== 'NotFoundException') {
```
bertyhell, line 313: err could be undefined
```suggestion
			if (err?.name !== 'NotFoundException') {
```

bertyhell (approved):

PR #382, opened 2023-04-11
Bugfix/arc 1581 return no empty fields

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1581


EmileVanTichelenSHD:
Closed because it seems to be working correctly without any code changes. Will reopen when there is more info about this bug.

PR #383, opened 2023-04-11
fix(arc-1606): Fix failing tests

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1606

### DIFF
--- a/src/modules/ie-objects/services/ie-objects.service.ts
+++ b/src/modules/ie-objects/services/ie-objects.service.ts
@@ -568,7 +568,9 @@ export class IeObjectsService {
 				...spaces.items.map((space) => space.maintainerId),
 				user.getMaintainerId(),
 			];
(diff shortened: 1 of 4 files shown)


tourlbr (approved):

src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -568,7 +568,8 @@ export class IeObjectsService {
 				...spaces.items.map((space) => space.maintainerId),
 				user.getMaintainerId(),
 			];
-		}
+		} else accessibleVisitorSpaceIds = visitorSpaceAccessInfo.visitorSpaceIds;
```
bertyhell, line 571: We don't use the short notation for ifs
```suggestion
		} else {
			accessibleVisitorSpaceIds = visitorSpaceAccessInfo.visitorSpaceIds;
		}
```

PR #384, opened 2023-04-12
Bugfix/arc 1608 kiosk no result

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1608

As vrt kiosk user, no objects are returned because there are no objects in db.
image

As Vlaams Parmlement kiosk user, you get 1 object (in postman)
image

### DIFF
--- a/src/modules/ie-objects/elasticsearch/queryBuilder.ts
+++ b/src/modules/ie-objects/elasticsearch/queryBuilder.ts
@@ -31,7 +31,7 @@ import {
 	VALUE_OPERATORS,
 } from './elasticsearch.consts';
 
-import { GroupId } from '~modules/users/types';
+import { GroupId, GroupName } from '~modules/users/types';
 import { PaginationHelper } from '~shared/helpers/pagination';
 import { SortDirection } from '~shared/types';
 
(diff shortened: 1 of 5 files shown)


src/modules/ie-objects/elasticsearch/queryBuilder.ts
```diff
@@ -197,16 +197,19 @@ export class QueryBuilder {
...
-		}
-
 		const filterObject: any = {};
 		// Add additional filters to the query object
 		const filterArray: any[] = [];
 		set(filterObject, 'bool.filter', filterArray);
 
+		// Kiosk users are only allowed to view objects from their maintainer
+		if (inputInfo.user.getGroupName() === GroupName.KIOSK_VISITOR) {
+			filterArray.push({
```
tourlbr, line 207: REQ: this will never work as you don't do anything with filterArray afterwards. You should check why logic in buildLicenseFilter doesn't work as intended.

```
// KIOSK users and CP Admins always have access to the visitor space that they are linked to.
		if (user.getMaintainerId()) {
```

tourlbr (changes requested):

src/modules/ie-objects/elasticsearch/queryBuilder.ts
EmileVanTichelenSHD: Resolved after decision to keep it like this after daily.

bertyhell (approved):

PR #386, opened 2023-04-17
fix(arc-1541): switch CM to bulk endpoint, fix newsletter check

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1541

* Use bulk import endpoint to avoid 429 rate limit errors
* Fix checking for subscribed to newsletter through custom field
* Don't clear values, but set them to be empty

The switch between on and off happens after 15 seconds in campaign monitor. So reloading the page after 15 seconds should reflect the change you made.

image
image

### DIFF
--- a/src/modules/campaign-monitor/services/campaign-monitor.service.ts
+++ b/src/modules/campaign-monitor/services/campaign-monitor.service.ts
@@ -187,7 +187,10 @@ export class CampaignMonitorService {
 			});
 
 			return {
-				newsletter: response?.State === 'Active',
+				newsletter:
+					response?.CustomFields.find(
+						(field) => field.Key === 'optin_mail_lists'
+					)?.Value?.includes('newsletter') ?? false,
 			};
 		} catch (err) {
 			if (err?.code === 203) {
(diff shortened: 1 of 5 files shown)


src/modules/campaign-monitor/services/campaign-monitor.service.ts
```diff
@@ -187,7 +187,10 @@ export class CampaignMonitorService {
 			});
 
 			return {
-				newsletter: response?.State === 'Active',
+				newsletter:
+					response?.CustomFields.find(
+						(field) => field.Key === 'optin_mail_lists'
```
SHDBramVandenbussche, line 192: Moet `'optin_mail_lists'` niet `'option_mail_lists'` zijn ?

SHDBramVandenbussche (approved):

silkederudder-shd (approved):

tourlbr (approved):

EmileVanTichelenSHD (changes requested):
campaign-monitor.service.spec.ts is failing. Please fix.
image

EmileVanTichelenSHD (approved):

PR #388, opened 2023-04-18
Feature/arc 1559 generate sitemap

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1559

Still a draft because some tests are still missing

### DIFF
--- a/src/modules/sitemap/controllers/sitemap.controller.ts
+++ b/src/modules/sitemap/controllers/sitemap.controller.ts
@@ -12,7 +12,8 @@ export class SitemapController {
 
 	@Post()
 	public async generateSitemap(): Promise<string> {
-		this.sitemapService.generateSitemap(); // no await because this can take a while and we don't want FE to crash
+		const sitemapConfig = await this.sitemapService.getSitemapConfig();
+		this.sitemapService.generateSitemap(sitemapConfig); // no await because this can take a while and we don't want FE to crash
 		return `${process.env.ASSET_SERVER_ENDPOINT}/${process.env.ASSET_SERVER_BUCKET_NAME}/${AssetFileType.SITEMAP}/index.xml`;
 	}
 }
--- a/src/modules/sitemap/services/sitemap.service.ts
+++ b/src/modules/sitemap/services/sitemap.service.ts
@@ -29,8 +29,7 @@ export class SitemapService {
 		protected assetsService: AssetsService
 	) {}
 
-	public async generateSitemap() {
-		const sitemapConfig = await this.getSitemapConfig();
+	public async generateSitemap(sitemapConfig: SitemapConfig) {
 		const contentPagesPaths = await this.getContentPagesPaths();
 
 		const staticPages = ['/', '/bezoek', '/zoeken'];
@@ -123,7 +122,11 @@ export class SitemapService {
 			const { app_config: config } = await this.dataService.execute<GetSitemapConfigQuery>(
 				GetSitemapConfigDocument
 			);
-
+			if (!config) {
+				throw new InternalServerErrorException(
+					'Database value for SITEMAP_CONFIG is not defined in app.config. Aborting Sitemap generation.'
+				);
+			}
 			return config[0];
 		} catch (err) {
 			throw new InternalServerErrorException('Failed getting sitemap config', err);
--- a/src/modules/sitemap/controllers/sitemap.controller.spec.ts
+++ b/src/modules/sitemap/controllers/sitemap.controller.spec.ts
@@ -1,5 +1,6 @@
 import { Test, TestingModule } from '@nestjs/testing';
 
+import { mockSitemapConfig } from '../mocks/sitemap.mocks';
 import { SitemapService } from '../services/sitemap.service';
 
 import { SitemapController } from './sitemap.controller';
@@ -9,6 +10,7 @@ import { TestingLogger } from '~shared/logging/test-logger';
 
 const mockSitemapService = {
 	generateSitemap: jest.fn(),
+	getSitemapConfig: jest.fn(),
 };
 
 describe('SitemapController', () => {
@@ -45,6 +47,9 @@ describe('SitemapController', () => {
 
 	describe('generateSitemap', () => {
 		it('should return the url of the index xml file', async () => {
+			mockSitemapService.getSitemapConfig.mockResolvedValueOnce({
+				app_config: [mockSitemapConfig],
+			});
 			mockSitemapService.generateSitemap.mockResolvedValueOnce('');
 
 			const result = await sitemapController.generateSitemap();
--- a/src/modules/sitemap/services/sitemap.service.spec.ts
+++ b/src/modules/sitemap/services/sitemap.service.spec.ts
@@ -144,9 +144,6 @@ describe('SitemapService', () => {
 
 	describe('generateSitemap', () => {
 		it('should return the xml for the general sitemap xml', async () => {
-			mockDataService.execute.mockResolvedValue({
-				app_config: [mockSitemapConfig],
-			});
 			mockContentPagesService.fetchContentPages.mockResolvedValueOnce([[mockContentPage], 1]);
 			mockSpacesService.findAll.mockResolvedValueOnce(mockSitemapSpaces);
 			mockIeObjectsService.findIeObjectsForSitemap.mockResolvedValueOnce({ total: 1 });
@@ -163,7 +160,7 @@ describe('SitemapService', () => {
 				'https://asset-server.be/bucketname/SITEMAP/index'
 			);
 
-			const result = await sitemapService.generateSitemap();
+			const result = await sitemapService.generateSitemap(mockSitemapConfig);
 			expect(result).toEqual(mockGeneralXml);
 		});
 	});


src/modules/assets/services/assets.service.ts
```diff
@@ -121,8 +121,8 @@ export class AssetsService {
 
 	public async upload(assetFiletype: AssetFileType, file: Express.Multer.File): Promise<string> {
 		const parsedFilename = path.parse(file.originalname);
-		const key = `${assetFiletype}/${_.kebabCase(parsedFilename.name)}-${uuidv4()}${
-			parsedFilename.ext
+		const key = `${assetFiletype}/${_.kebabCase(parsedFilename.name)}${
+			assetFiletype === AssetFileType.SITEMAP ? '' : `-${uuidv4()}${parsedFilename.ext}`
```
bertyhell, line 125:
```
public async upload(assetFiletype: AssetFileType, file: Express.Multer.File, prefferedKey?: string): Promise<string> {
```

```suggestion
		const key = prefferedKey ?? `${assetFiletype}/${_.kebabCase(parsedFilename.name)}-${uuidv4()}${
			parsedFilename.ext
		}`;
```

src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -275,6 +284,23 @@ export class IeObjectsService {
 		return allAdapted;
 	}
 
+	public async findObjectsForSitemap(licenses: IeObjectLicense[]): Promise<IeObjectsSitemap[]> {
+		try {
+			const { object_ie: objects } = await this.dataService.execute<
+				FindObjectsForSitemapQuery,
+				FindObjectsForSitemapQueryVariables
+			>(FindObjectsForSitemapDocument, { licenses });
```
bertyhell, line 292: this will get all 1.4 million objects at once from the database.
I think we better do this like:
```typescript
findObjectsForSitemap(licenses: IeObjectLicense[], offset: number, limit: number): Promise<IPagination<IeObjectsSitemap>> {
   ...
}

const result = findObjectsForSitemap(..., 0, 0);
for (let i = 0; i < result.total; i+=XML_OBJECTS_SIZE) {
	const objects = await findObjectsForSitemap(..., i, i+XML_OBJECTS_SIZE);

	const xmlSitemapString = convertToXml(objects.items);
	await uploadFile(...xmlSitemapString...)
}
```

So we handle 50000 objects at a time, and dont overload the database with one giant request.

src/modules/sitemap/controllers/sitemap.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+	@Post()
+	public async generateSitemap(): Promise<string> {
+		/*	From Jira:	// Comment can be removed once this feature is complete
+					items in sitemap:
+						-homepage (static): '/' done
+						-landingspagina bzt (static): '/bezoek' done
+						-andere contentpagina’s: '' admincore : fetchContentPages //like: import { DataService } from '@meemoo/admin-core-api'; done
+						-zoekpagina algemeen (static): '/zoeken' done
+						-zoekpagina per cp: '/zoeken/?aanbieders={or id van aanbieder}' done
+						-item-detailpagina’s: '/zoeken/{maintainerSlug}/{objectSchemaId}' done
```
bertyhell, line 22:
```suggestion
						-item-detailpagina’s: '/zoeken/{maintainerSlug}/{objectSchemaId}/{objectTitle}' done
```

src/modules/sitemap/services/sitemap.service.ts
```diff
@@ -0,0 +1,207 @@
...
+				})),
+				...contentPagesPaths.map((path) => ({
+					loc: process.env.CLIENT_HOST + path,
+					changefreq: 'monthly',
+				})),
+				// '/zoeken' is seperate because of the required order
+				{
+					loc: process.env.CLIENT_HOST + '/zoeken',
+					changefreq: 'monthly',
+				},
```
bertyhell, line 61: i don't think this is a hard requirement to have search after the other content pages

src/modules/sitemap/services/sitemap.service.ts
```diff
@@ -0,0 +1,207 @@
...
+				...activeSpaces.items.map((space) => ({
+					loc: process.env.CLIENT_HOST + '/zoeken/?aanbieders=' + space.maintainerId,
+					changefreq: 'weekly',
+					lastmod: moment(space.updatedAt).format('YYYY-MM-DD'),
```
bertyhell, line 65: i wouldn't include this lastmod date for the spaces, since objects can be added to a space without the space being modified. For objects this is good to add, but not for spaces.

src/modules/sitemap/services/sitemap.service.ts
```diff
@@ -0,0 +1,207 @@
...
+			if (configPaths.includes(path)) {
+				const configValue = config.value.find((c) => c.path === path);
+				if (configValue?.blacklisted === true) {
+					return result;
+				} else {
+					return [...result, { ...page, priority: configValue?.priority }];
+				}
+			}
+			return [...result, page];
+		}, []);
```
bertyhell, line 205: I'm not a big fan of reduce since i find it hard to see what is going on. I prefer to use compact(....map(...))
compact removes all null/undefined values from an array

you don't need to compare with true inside an if statement

We can also easily set the priority on the page even though it wasn't a requirement.

statements like this: `[...result, page];` can get quite expensive when you use them in a for loop and on large arrays (50000). Since every loop you're asking the pc to copy all values into a new array. Which causes a lot of memory copying. After the reduce you'll have copied 2.500.000.000 page items. 50.000 * 50.000

```suggestion
		return compact(pages.map((page) => {
			const path = page.loc.substring(process.env.CLIENT_HOST.length);
			if (configPaths.includes(path)) {
				const configValue = config.value.find((c) => c.path === path);
				if (configValue?.blacklisted) {
					return null;
				}
				if (configValue?.priority) {
					page.priority = configValue?.priority;
				}
			}
			return page;
		}));
```

bertyhell (changes requested):

src/modules/sitemap/services/sitemap.service.ts
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	private blacklistAndPrioritizePages(
+		pages: SitemapItemInfo[],
+		config: SitemapConfig
+	): SitemapItemInfo[] {
+		const configPaths = config.value.map((c) => c.path);
+
+		return pages.reduce((result, page) => {
+			const path = page.loc.substring(process.env.CLIENT_HOST.length);
```
bertyhell, line 195: OPTIONAL: it might be easier to append the process.env.CLIENT_HOST to the urls in the renderPage function. So you can compare paths easily until you convert the links to xml.

src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -275,6 +284,23 @@ export class IeObjectsService {
 		return allAdapted;
 	}
 
+	public async findObjectsForSitemap(licenses: IeObjectLicense[]): Promise<IeObjectsSitemap[]> {
+		try {
+			const { object_ie: objects } = await this.dataService.execute<
```
tourlbr, line 289: REQ: rename objects to u ieObjects. This to adhere to naming conventions and avoid confusion with Javascripts own object.

src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -275,6 +284,23 @@ export class IeObjectsService {
 		return allAdapted;
 	}
 
+	public async findObjectsForSitemap(licenses: IeObjectLicense[]): Promise<IeObjectsSitemap[]> {
```
tourlbr, line 287: REQ: rename findObjectsForSitemap to **findIeObjectsForSitemap**. Everything is called IeObject and not just object for a reason.

src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -275,6 +284,23 @@ export class IeObjectsService {
...
+		try {
+			const { object_ie: objects } = await this.dataService.execute<
+				FindObjectsForSitemapQuery,
```
tourlbr, line 290: REQ: **FindIeObjectsForSitemapQuery**

src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -275,6 +284,23 @@ export class IeObjectsService {
...
+				FindObjectsForSitemapQueryVariables
```
tourlbr, line 291: REQ: **FindIeObjectsForSitemapQueryVariables**

src/modules/ie-objects/services/ie-objects.service.ts
tourlbr, line 292: REQ: **FindIeObjectsForSitemapDocument**

src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -275,6 +284,23 @@ export class IeObjectsService {
...
+			>(FindObjectsForSitemapDocument, { licenses });
+
+			const adapted = objects.map((object) => {
```
tourlbr, line 294: REQ: rename adapted to adaptedIeObjects

src/modules/sitemap/services/sitemap.service.ts
```diff
@@ -0,0 +1,207 @@
...
+		const contentPagesPaths = await this.getContentPagesPaths();
+
+		const staticPages = ['/', '/bezoek', '/geheime-content-pagina']; // /geheime-content-pagina is for testing
+
+		const activeSpaces = await this.spacesService.findAll(
+			{ size: 1000, status: [VisitorSpaceStatus.Active] },
+			null
+		);
+
+		const publicObjects = await this.ieObjectsService.findObjectsForSitemap([
```
tourlbr, line 41: REQ: rename **publicIeObjects**

src/modules/sitemap/services/sitemap.service.ts
```diff
@@ -0,0 +1,207 @@
...
+					lastmod: moment(space.updatedAt).format('YYYY-MM-DD'),
+				})),
+			],
+			sitemapConfig
+		);
+
+		// Create sitemap info for item detail pages
+		const itemDetailPages: SitemapItemInfo[] = this.blacklistAndPrioritizePages(
+			[
+				...publicObjects.map((object) => ({
```
tourlbr, line 74: REQ: rename **publicIeObjects**

src/modules/sitemap/services/sitemap.service.ts
```diff
@@ -0,0 +1,207 @@
...
+						object.maintainerSlug +
+						'/' +
+						object.schemaIdentifier,
+					changefreq: 'weekly',
+					lastmod: moment(object.updatedAt).format('YYYY-MM-DD'),
+				})),
+			],
+			sitemapConfig
+		);
+		const XML_OBJECTS_SIZE = 50000; // amount of objects in 1 xml file, should be 50000
```
tourlbr, line 87: REQ: consts like these should be in a consts file so that they can be easily edited in one place.

src/modules/sitemap/sitemap.types.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export interface SitemapConfig {
+	value: PathInfo[];
+	name: string;
+	updated_at?: string;
+	created_at?: string;
+}
+
+export interface PathInfo {
```
tourlbr, line 15: REQ: rename **SitemapPathInfo**. Please use the provided naming conventions as used in other modules types. Always use name of the module + append according to the implementation.

Usecase: what if we define PathInfo in another module and we have to use both in the same module when imported? Then we would have a problem (yes we can alias it but its better to do it right the first time).

src/modules/ie-objects/queries/find-objects-for-sitemap.graphql
```diff
@@ -0,0 +1,12 @@
+query findObjectsForSitemap($licenses: [String!]) {
```
tourlbr, line 1: REQ: rename findIeObjectsForSitemap

tourlbr (changes requested):

src/modules/assets/services/assets.service.ts
```diff
@@ -119,10 +119,16 @@ export class AssetsService {
...
+		assetFiletype: AssetFileType,
+		file: Express.Multer.File,
+		prefferedKey?: string
+	): Promise<string> {
 		const parsedFilename = path.parse(file.originalname);
-		const key = `${assetFiletype}/${_.kebabCase(parsedFilename.name)}-${uuidv4()}${
-			parsedFilename.ext
+		const key = `${assetFiletype}/${
+			prefferedKey
+				? _.kebabCase(prefferedKey)
```
bertyhell, line 130: i would move the kebabCase function to where you call upload
otherwise you're modifying the prefferedKey evan though the consumer of the upload function might want an exact key that will not be modified.

src/modules/ie-objects/services/ie-objects.service.spec.ts
```diff
@@ -269,6 +271,36 @@ describe('ieObjectsService', () => {
...
 
+	describe('findObjectsForSitemap', () => {
+		it('should throw an error when it fails to get object', async () => {
+			mockDataService.execute.mockResolvedValueOnce('');
+			try {
+				await ieObjectsService.findIeObjectsForSitemap(
+					[IeObjectLicense.PUBLIEK_METADATA_LTD, IeObjectLicense.PUBLIEK_METADATA_ALL],
+					0,
+					50
+				);
```
bertyhell, line 282: add a fail statement after this await, since we want the test to fail if the function does not throw an exception

src/modules/sitemap/controllers/sitemap.controller.ts
```diff
@@ -0,0 +1,27 @@
...
+		*/
+		this.sitemapService.generateSitemap(); // no await because this can take a while and we don't want FE to crash
+		return `${process.env.ASSET_SERVER_ENDPOINT}/${process.env.ASSET_SERVER_BUCKET_NAME}/${AssetFileType.SITEMAP}/index`;
```
bertyhell, line 25: OPTIONAL: maybe we should set the filename to index.xml, so someone looking at the url will know what format to expect without having to load the file.

src/modules/sitemap/services/sitemap.service.ts
```diff
@@ -0,0 +1,205 @@
...
+		protected contentPagesService: ContentPagesService,
+		protected ieObjectsService: IeObjectsService,
+		protected assetsService: AssetsService
+	) {}
+
+	public async generateSitemap() {
+		const sitemapConfig = await this.getSitemapConfig();
+		const contentPagesPaths = await this.getContentPagesPaths();
+
+		const staticPages = ['/', '/bezoek', '/zoeken', '/geheime-content-pagina']; // /geheime-content-pagina is for testing
```
bertyhell, line 36: we should remove /geheime-content-pagina since we can check blacklisting in the tests
if needed you could also do:

```
public async generateSitemap(staticPages: string[]) {
```

so you can pass the value from the controller and from the test

src/modules/sitemap/services/sitemap.service.ts
```diff
@@ -0,0 +1,205 @@
...
+				destination: '',
+				filename: name,
+				path: '',
+				buffer: Buffer.from(xml, 'utf-8'),
+			},
+			name
+		);
+	}
+
+	private blacklistAndPrioritizePages(
```
bertyhell, line 185: i don't see any tests for this function

src/modules/sitemap/services/sitemap.service.ts
```diff
@@ -0,0 +1,205 @@
...
+	private renderPage(pageInfo: SitemapItemInfo): string {
+		return `<url>
+			<loc>${process.env.CLIENT_HOST}${pageInfo.loc}</loc>
+			${pageInfo.lastmod ? `<lastmod>${pageInfo.lastmod}</lastmod>` : ``}
+			<changefreq>${pageInfo.changefreq}</changefreq>
+			${pageInfo.priority ? `<priority>${pageInfo.priority}</priority>` : ``}
+			</url>`;
+	}
+
+	private renderIndexPage(url: string): string {
```
bertyhell, line 160: doesn't this name make more sense?
renderSitemapIndexEntry

bertyhell (approved):

tourlbr (approved):

PR #397, opened 2023-04-25
bugfix/ARC-1165: Several material request fixes

EmileVanTichelenSHD:
Bugs are in comment of https://meemoo.atlassian.net/browse/ARC-1165

Tickets:
https://meemoo.atlassian.net/browse/ARC-1631
https://meemoo.atlassian.net/browse/ARC-1634
https://meemoo.atlassian.net/browse/ARC-1643

ARC-1633 is not in this PR because we are waiting on a response.

Please test together with FE PR: https://github.com/viaacode/hetarchief-client/pull/818


src/modules/material-requests/services/material-requests.service.ts
```diff
@@ -204,17 +204,19 @@ export class MaterialRequestsService {
...
+		const { type, reason, organisation, requester_capacity, is_pending, updated_at } =
+			materialRequestInfo;
 
 		const updateMaterialRequest = {
 			type,
 			reason,
 			organisation,
 			requester_capacity,
 			is_pending,
+			updated_at,
```
tourlbr, line 219: It seems strange to me that you would receive a value **updated_at** from the FE. In most use cases you retrieve all the changing data from FE and set a new value for **updated_at** in the BE.

REQ: remove option to retrieve **updated_at**

tourlbr (changes requested):

src/modules/material-requests/services/material-requests.service.ts
EmileVanTichelenSHD: The value is set in the controller. We can also update the 'updated_at ' in the service but that's not what is requested in the ticket.

src/modules/material-requests/dto/material-requests.dto.ts
```diff
@@ -132,14 +133,14 @@ export class CreateMaterialRequestDto {
...
 		type: String,
 		description: "The reason for this user's material request",
 		example:
 			'I would like to do research on evolution of the Dutch language in the vrt news across the decades.',
 	})
-	reason: string;
+	reason?: string = '';
 
 	@IsString()
 	@IsEnum(MaterialRequestRequesterCapacity)
```
bertyhell, line 146: We should delete the homemade enum and replace it with the database generated enum everywhere
MaterialRequestRequesterCapacity => Lookup_App_Material_Request_Requester_Capacity_Enum

bertyhell (approved):

PR #400, opened 2023-04-27
fix(arc-1651): fix searching for terms starting with digit

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1651
Note: the pid in the ticket '3r0pr9jj7g' doesn't seem to exist in ES.

Before:
image
image
image

After:
image
image
image

### DIFF
--- a/src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
+++ b/src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
@@ -202,3 +202,5 @@ export const NEEDS_AGG_SUFFIX: { [prop in IeObjectsSearchFilterField]?: string }
 	[IeObjectsSearchFilterField.GENRE]: 'keyword',
 	[IeObjectsSearchFilterField.TYPE]: 'keyword',
 };
+
+export const ES_DIGIT_PREFIX = '_ç_ç_ç_ç_ç_';
--- a/src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
+++ b/src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
@@ -206,7 +206,6 @@ export const NEEDS_AGG_SUFFIX: { [prop in IeObjectsSearchFilterField]?: string }
 /*
 	This prefix is added before and after every digit before jsep parses it.
 	This has to be a string that's not expected to occur in any ES object's properties.
-	It should also be made of characters that aren't recognised by jsep.
 	If the searchterm is a string starting with a digit and we don't do this, jsep will throw an error saying a variable cannot start with a digit.
 */
-export const ES_DIGIT_PREFIX = '_ç_ç_ç_ç_ç_';
+export const JSEP_DIGIT_PREFIX = '_ç_ç_ç_ç_ç_';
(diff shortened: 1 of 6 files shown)


src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
```diff
@@ -202,3 +202,5 @@ export const NEEDS_AGG_SUFFIX: { [prop in IeObjectsSearchFilterField]?: string }
 	[IeObjectsSearchFilterField.GENRE]: 'keyword',
 	[IeObjectsSearchFilterField.TYPE]: 'keyword',
 };
+
+export const ES_DIGIT_PREFIX = '_ç_ç_ç_ç_ç_';
```
tourlbr, line 206: REQ: please add some comments to further elaborate why this is needed and what it is. For future developers this is just something weird.

src/modules/ie-objects/helpers/encode-search-term.ts
```diff
@@ -0,0 +1,10 @@
+import { ES_DIGIT_PREFIX } from '../elasticsearch/elasticsearch.consts';
+
```
tourlbr, line 2: REQ: please add some comments to further elaborate what is being done here (+ some examples (not tests) but hardcoded examples).

src/modules/ie-objects/helpers/convert-node-to-es-query-filter-objects.ts
```diff
@@ -3,11 +3,14 @@ import { baseTypes, Expression } from 'jsep';
...
 import { SearchFilter } from '../dto/ie-objects.dto';
 
+import { decodeSearchterm } from './encode-search-term';
+
 export const convertNodeToEsQueryFilterObjects = (
 	node: Expression,
 	searchTemplates?: { fuzzy: any[]; exact: any[] },
 	searchFilter?: SearchFilter
 ): any => {
+	node.name = decodeSearchterm(node.name as string);
```
tourlbr, line 13: REQ: add newline after this

tourlbr (changes requested):
Always make sure when you add new code it is clear what it is or what it is going to do. Usually with clear variable and function naming this is enough. But for specific use case it is not enough and some elaboration is needed.

tourlbr (approved):

src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
```diff
@@ -202,3 +202,11 @@ export const NEEDS_AGG_SUFFIX: { [prop in IeObjectsSearchFilterField]?: string }
 	[IeObjectsSearchFilterField.GENRE]: 'keyword',
 	[IeObjectsSearchFilterField.TYPE]: 'keyword',
 };
+
+/*
+	This prefix is added before and after every digit before jsep parses it.
+	This has to be a string that's not expected to occur in any ES object's properties.
+	It should also be made of characters that aren't recognised by jsep.
```
bertyhell, line 209: I would omit this line. It causes more confusion that it adds explanation
```suggestion
```

src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
```diff
@@ -202,3 +202,11 @@ export const NEEDS_AGG_SUFFIX: { [prop in IeObjectsSearchFilterField]?: string }
...
+	If the searchterm is a string starting with a digit and we don't do this, jsep will throw an error saying a variable cannot start with a digit.
+*/
+export const ES_DIGIT_PREFIX = '_ç_ç_ç_ç_ç_';
```
bertyhell, line 212: This has nothing to do with elasticsearch. It is only used before and after the parsing with jsep
```suggestion
export const JSEP_DIGIT_PREFIX = '_ç_ç_ç_ç_ç_';
```

bertyhell (approved):

PR #412, opened 2023-05-15
Bugfix/arc 1686 search with special character

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1686

Before:
image

After:
image

### DIFF
--- a/src/modules/ie-objects/helpers/convert-string-to-search-terms.ts
+++ b/src/modules/ie-objects/helpers/convert-string-to-search-terms.ts
@@ -19,7 +19,10 @@ export function convertStringToSearchTerms(searchQuery: string): string[] {
 }
 
 export function convertNodeToSearchTerms(node: Expression): string[] {
+	node.raw = decodeSearchterm(node.raw as string);
+	node.value = decodeSearchterm(node.value as string);
 	node.name = decodeSearchterm(node.name as string);
+
 	switch (node.type) {
 		case 'Compound':
 			return ((node.body || []) as Expression[]).flatMap(convertNodeToSearchTerms);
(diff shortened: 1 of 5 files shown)


src/modules/ie-objects/helpers/convert-string-to-search-terms.ts
```diff
@@ -19,7 +19,10 @@ export function convertStringToSearchTerms(searchQuery: string): string[] {
 }
 
 export function convertNodeToSearchTerms(node: Expression): string[] {
```
bertyhell, line 21: we should check that this function also returns the whole literal string als a term in a test case:
```
		filters: [
			{
				field: IeObjectsSearchFilterField.QUERY,
				operator: Operator.CONTAINS,
				value: 'De grote -)( test',
			},
		],		
```

response should contain:
```
{
	...
	searchTerms: ["De grote -)( test"],
}
```

bertyhell (changes requested):

src/modules/ie-objects/helpers/convert-string-to-search-terms.ts
```diff
@@ -19,7 +19,10 @@ export function convertStringToSearchTerms(searchQuery: string): string[] {
...
+	node.raw = decodeSearchterm(node.raw as string);
```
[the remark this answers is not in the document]
EmileVanTichelenSHD: After some checking i removed it.

bertyhell (approved):

src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -113,11 +114,26 @@ export class IeObjectsService {
...
+		let esQuery;
+		try {
+			esQuery = QueryBuilder.build(inputQuery, {
+				user,
+				visitorSpaceInfo,
+				spacesIds,
+			});
+		} catch (err) {
+			/*
+					If the QueryBuilder throws an error, we try the query with a literal string.
```
tourlbr, line 126: REQ: fix indentiation issue

tourlbr (approved):

PR #414, opened 2023-05-16
fix(arc-1640): fix sending request list if no contactMail

EmileVanTichelenSHD:
https://meemoo.atlassian.net/browse/ARC-1640

The code would throw an error if the maintainer had no contact info.

### DIFF
--- a/src/modules/material-requests/controllers/material-requests.controller.ts
+++ b/src/modules/material-requests/controllers/material-requests.controller.ts
@@ -170,6 +170,7 @@ export class MaterialRequestsController {
 			});
 
 			materialRequests.items.forEach((materialRequest: MaterialRequest) => {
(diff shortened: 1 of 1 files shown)


src/modules/material-requests/controllers/material-requests.controller.ts
```diff
@@ -170,7 +170,7 @@ export class MaterialRequestsController {
 			});
 
 			materialRequests.items.forEach((materialRequest: MaterialRequest) => {
-				materialRequest.contactMail = materialRequest.contactMail.find(
+				materialRequest.contactMail = materialRequest?.contactMail?.find(
```
bertyhell, line 173: maybe add a comment:
```suggestion
				// If the email does not exist, the campaign monitor service will default to process.env.MEEMOO_MAINTAINER_MISSING_EMAIL_FALLBACK
				materialRequest.contactMail = materialRequest?.contactMail?.find(
```

bertyhell (approved):

tourlbr (approved):

PR #440, opened 2023-06-27
fix(arc-1834): do not show object with license if no keyuser

WardVerc1:
Ticket: https://meemoo.atlassian.net/browse/ARC-1834

### DIFF
--- a/src/modules/collections/controllers/collections.controller.ts
+++ b/src/modules/collections/controllers/collections.controller.ts
@@ -108,28 +108,23 @@ export class CollectionsController {
 
 		const isKeyUser = user.getIsKeyUser();
 
-		// Limit access to the objects in the collection
+		// Limit access to the objects in the collection:
+		// https://meemoo.atlassian.net/browse/ARC-1834
 		folderObjects.items = (folderObjects.items ?? [])
 			.filter((object) => {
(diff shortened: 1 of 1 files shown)


SHDBramVandenbussche (approved):

src/modules/collections/controllers/collections.controller.ts
```diff
@@ -106,10 +106,39 @@ export class CollectionsController {
...
+					object.licenses &&
+					object.licenses.length === 2 &&
+					object.licenses.includes(
+						IeObjectLicense.INTRA_CP_CONTENT && IeObjectLicense.INTRA_CP_METADATA_ALL
+					);
+				if (noKeyUserAndHasOnlyLicense || noKeyUserAndHasLicenses) {
+					return;
+				} else {
+					return object;
+				}
```
bertyhell, line 133: i think this does the same thing:
Also best to always return only booleans in a filter function, not undefined or objects
```suggestion
				const hasOnlyIntraCpLicenses =
					(object.licenses || []).filter(license => 
						![IeObjectLicense.INTRA_CP_CONTENT, IeObjectLicense.INTRA_CP_METADATA_ALL].includes(license)
					).length === 0;

				if (!isKeyUser && hasOnlyIntraCpLicenses) {
					return false;
				} else {
					return true;
				}
```

bertyhell (changes requested):

bertyhell (approved):

PR #472, opened 2024-05-17
feat: add mockdata endpoint

RobbeBierebeeck:


src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -90,6 +90,11 @@ export class IeObjectsController {
 		);
 	}
 
+	@Get('newspaper-titles')
+	public async getNewspaperTitles() {
+		return this.ieObjectsService.getNewspaperTitles();
+	}
```
bertyhell, line 96: for documentation purposes, also specify the response type here
```suggestion
	@Get('newspaper-titles')
	public async getNewspaperTitles(): Promise<NewspaperTitle[]> {
		return this.ieObjectsService.getNewspaperTitles();
	}
```

src/modules/ie-objects/ie-objects.types.ts
```diff
@@ -313,6 +313,10 @@ export interface IeObjectsSitemap {
 	updatedAt: string;
 }
 
+export interface InewspaperTitle {
```
bertyhell, line 316: We don't follow the I for interface convention in these repos
```suggestion
export interface NewspaperTitle {
```

image

src/shared/i18n/locales/nl.json
```diff
@@ -1,44 +1,44 @@
 {
-  "modules/auth/controllers/het-archief___account-configuratie": "Account configuratie",
-  "modules/auth/controllers/meemoo___account-configuratie": "Account configuratie",
-  "modules/auth/services/idp___account-configuratie": "Account configuratie",
-  "modules/auth/services/idp___de-account-is-een-beheerder-maar-heeft-geen-organisatie-in-de-acm-voeg-een-organisatie-toe-in-de-acm-no-org-linked": "Er is een probleem met deze admin account: er is geen organisatie geregistreerd. Neem contact op met <a href=\"mailto:support@meemoo.be\" target=\"\">support@meemoo.be</a> en laat even weten welk e-mailadres je gebruikt om in te loggen. We brengen het zo snel mogelijk in orde.",
-  "modules/auth/services/idp___de-account-is-een-kiosk-gebruiker-maar-heeft-geen-organisatie-in-de-acm-voeg-een-organisatie-toe-in-de-acm-no-org-linked": "Er is een probleem met deze kiosk-account: er is geen organisatie geregistreerd. Neem contact op met <a href=\"mailto:support@meemoo.be\" target=\"\">support@meemoo.be</a> en laat even weten welk e-mailadres je gebruikt om in te loggen. We brengen het zo snel mogelijk in orde.",
```
bertyhell, line 6: all these translations are english
they should notmally be dutch

can you check:
* that you're on the latest commit of the hetarchief hasura
* that the localhost:9000 app.translations table contains both english and dutch translations and the "language" column contains both EN and NL?
* If not, run hasura seed apply again
* and then admin-core-ui: npm run extract hetarchief translations

bertyhell (changes requested):

PR #473, opened 2024-05-24
feat(ARC-2141): add ie-objects endpoint

RobbeBierebeeck:


src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -343,28 +343,31 @@ export class IeObjectsService {
...
-		}
-
-		const adapted = this.adaptFromDB(objectIe[0]);
-		adapted.thumbnailUrl = await this.playerTicketService.resolveThumbnailUrl(
-			adapted.thumbnailUrl,
-			referer,
-			ip
+		// if (!objectIe[0]) {
+		// 	throw new NotFoundException(`Object IE with id '${schemaIdentifier}' not found`);
+		// }
```
bertyhell, line 359: move this to the wrapping function, so the 404 exception in the get single ie-object stays the same
```suggestion
```

src/modules/collections/controllers/collections.controller.ts
```diff
@@ -229,10 +229,10 @@ export class CollectionsController {
 
 		// Log event
 		const ieObject = await this.ieObjectsService.findBySchemaIdentifier(
-			objectSchemaIdentifier,
+			[objectSchemaIdentifier],
 			referer,
 			getIpFromRequest(request)
-		);
+		)[0];
```
bertyhell, line 235: make sure you await the right thing
you want to await the return of the function, not the promise[0]
(await this.ieObjectsService.findBySchemaIdentifier())[0]

bertyhell (approved):

src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -441,4 +441,51 @@ export class IeObjectsController {
...
+		const ieObject: IeObject[] = await this.ieObjectsService.findBySchemaIdentifier(
+			ids,
+			referer,
+			getIpFromRequest(request)
+		);
+
+		const visitorSpaceAccessInfo =
+			await this.ieObjectsService.getVisitorSpaceAccessInfoFromUser(user);
+
+		return ieObject.map((ieObject) => {
```
bertyhell, line 461: don't use variable names in for loops or map functions that already exist outside of the loop

src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -441,4 +441,51 @@ export class IeObjectsController {
...
+		@Query('id') ids: string[],
+		@Headers('referer') referer: string,
+		@Req() request: Request,
+		@SessionUser() user: SessionUserEntity
+	): Promise<IeObject[] | Partial<IeObject>[]> {
+		const ieObject: IeObject[] = await this.ieObjectsService.findBySchemaIdentifier(
+			ids,
+			referer,
+			getIpFromRequest(request)
+		);
```
bertyhell, line 456:
```suggestion
		const ieObjects: IeObject[] = await this.ieObjectsService.findBySchemaIdentifier(
			ids,
			referer,
			getIpFromRequest(request)
		);
		const ieObject = ieObjects[0];
```

bertyhell (approved):

PR #476, opened 2024-05-29
feat(ARC-2082): add endpoint so user language settings can be updated

RobbeBierebeeck:


src/modules/users/dto/users.dto.ts
```diff
@@ -57,3 +57,8 @@ export class UpdateAcceptedTosDto {
 	})
 	acceptedTosAt: string;
 }
+
+export class UpdateUserLangDto {
+	@IsIn(['nl', 'en'])
```
bertyhell, line 62: usually we type them like this. We want to avoid repeating the 'nl' and 'en' as much as possible, since if another language is added we need to change it in a million places
```
	@IsString()
	@ApiProperty({
		type: String,
		enum: LanguageCode,
	})
	language: LanguageCode;
```

bertyhell (approved):

PR #489, opened 2024-07-11
feat(ARC-2144): get newspapertitles from db

RobbeBierebeeck:


src/modules/ie-objects/queries/getNewspaperTitles.graphql
```diff
@@ -0,0 +1,11 @@
...
+	users_permission_aggregate {
+		aggregate {
+			count
+		}
+	}
+	graph__newspapers_public {
+		id
+		schema_name
+	}
+}
```
bertyhell, line 11: i suspect you didn't want to fetch the user permission count in this query
```suggestion
	graph__newspapers_public {
		id
		schema_name
	}
}
```

bertyhell (approved):

PR #491, opened 2024-07-22
ARC-2168 Add endpoint to download full newspaper or single page with metadata and alto xml

bertyhell:
https://meemoo.atlassian.net/browse/ARC-2168

image
image


package.json
```diff
@@ -94,6 +94,7 @@
     "@nestjs/schematics": "9.2.0",
     "@nestjs/testing": "9.4.3",
     "@studiohyperdrive/commitlint-config": "^2.2.0",
+    "@types/archiver": "^6.0.2",
```
RobbeBierebeeck, line 97: you use here a ^ to define the version. I see there are other dev dependencies that do not have this carrot and use fixed version. Do we have a convention around this or why would we switch between the two?

src/modules/newspapers/controllers/newspapers.controller.ts
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		const zipEntries: { filename: string; type: 'url' | 'string'; value: string }[] = [];
+
+		// Extract images and alto xml urls from the object
+		const exportSinglePage = !isNil(pageIndex) && !isNaN(pageIndex);
+		const pagesToExport = exportSinglePage
+			? limitedObjectMetadata.pageRepresentations.slice(pageIndex, pageIndex + 1)
+			: limitedObjectMetadata.pageRepresentations;
+		pagesToExport.forEach((representations, pageRepresentationIndex) => {
```
RobbeBierebeeck, line 74: Can you add more comments or maybe split it up in some more chucks? The 3 nested forEach loops make it really hard to read.

RobbeBierebeeck (approved):

package.json
bertyhell: 👍 fixed in commit on develop

src/modules/newspapers/controllers/newspapers.controller.ts
bertyhell: 👍 fixed in commit on develop

PR #495, opened 2024-07-26
feat(ARC-2194): add related meemooid endpoint

RobbeBierebeeck:


src/modules/ie-objects/controllers/ie-objects.controller.ts
```diff
@@ -238,6 +238,16 @@ export class IeObjectsController {
...
 	}
 
+	@Get('schemaIdentifierLookup/:schemaIdentifier')
+	@ApiOperation({
+		description: 'Get new schemaIdentifier when given the old one as param.',
+	})
+	public async identifierLookup(
+		@Param('schemaIdentifier') schemaIdentifier: string
+	): Promise<string> {
+		return this.ieObjectsService.getRelatedIdentifier(schemaIdentifier);
```
bertyhell, line 248: we already have related objects as a concept
so i would prefer to not use that naming here
maybe?
```suggestion
		return this.ieObjectsService.getSchemaIdentifierV3(schemaIdentifierV2);
```

bertyhell (approved):

PR #508, opened 2024-08-29
fix(ARC-2473): add event type

RobbeBierebeeck:


bertyhell:
did you verify that the events are now correctly being entered into the logging database?

bertyhell (approved):

PR #614, opened 2026-03-24
ARC-3498: return event history of request with request

reunefe:


src/modules/material-request-messages/services/material-request-messages.service.ts
```diff
@@ -74,18 +75,27 @@ export class MaterialRequestMessagesService {
...
+	public adaptEvent(
 		message: GetMaterialRequestMessagesQuery['app_material_request_messages_and_events'][0]
-	): MaterialRequestMessage {
+	): MaterialRequestEvent {
 		return {
 			id: message.id,
 			materialRequestId: message.material_request_id,
+			messageType: message.message_type,
+			body: message.body,
+			createdAt: message.created_at,
```
bertyhell, line 86: for an event the client also wants to know who sent it. doesn't it make more sense to use a single interface for both events and messages and just have events with attachementUrl and attachementFilename: null?

so use a single adapt function for both messages aand events
or is the adaptEvent only used when the client only fetches the events without messages

bertyhell (approved):

src/modules/material-request-messages/services/material-request-messages.service.ts
reunefe: The adaptEvent is only for the history, not for the chat. The adapt function for the chat will include all that data and builds upon the adaptEvent. We could align them, but I'm not sure if that is really needed

PR #617, opened 2026-03-30
feat(ARC-3510): add sorting to documents endpoint

bertyhell:
https://meemoo.atlassian.net/browse/ARC-3510

image


src/modules/material-request-messages/dto/material-request-messages.dto.ts
```diff
@@ -0,0 +1,51 @@
...
+	})
+	page? = 1;
+
+	@IsNumber()
+	@Type(() => Number)
+	@IsOptional()
+	@ApiPropertyOptional({
+		type: Number,
+		description: 'The max. number of results to return',
+		default: 10,
```
reunefe: The FA speaks of max 20, should that not be the default?

reunefe (approved):

PR #631, opened 2026-04-20
Bugfix: Ensure attachments to be always in the same order

reunefe:


src/modules/material-request-messages/services/material-request-messages.service.ts
```diff
@@ -130,7 +130,12 @@ export class MaterialRequestMessagesService {
 	): MaterialRequestMessage {
 		return {
 			...this.adaptEvent(message),
-			attachments: message.attachments?.map((attachment) => this.adaptAttachment(attachment)) || [],
+			attachments: (message.attachments || [])
+				.map((attachment) => this.adaptAttachment(attachment))
+				.sort(
+					(first, second) =>
+						new Date(first.createdAt).getTime() - new Date(second.createdAt).getTime()
+				),
```
bertyhell: doesn't it make more sense to do this sorting in the database query?
```graphql
		attachments(order_by: { created_at: asc }) {
			id
			attachment_url
			attachment_filename
			created_at
		}
```

bertyhell (approved):

src/modules/material-request-messages/services/material-request-messages.service.ts
reunefe: Yes it does, But I couldn't find how xD

bertyhell (approved):

PR #636, opened 2026-04-22
feat(ie-objects): simplify ie object fetch from database

bertyhell:
should make it easier to get a database result from production and debug it locally in a test since the whole ie-object is fetched in a single query which should be more efficient than doing a promise.all of multiple smaller sub queries


src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -621,60 +592,56 @@ export class IeObjectsService {
...
+		const hasPartResponse = ieObjectResponse.getHasPart;
+		const isRepresentedByResponse = ieObjectResponse.getIsRepresentedBy;
+
+		if (!ie) {
+			return null;
+		}
 
 		const ieObjectId = ie?.id;
-		if (!ieObjectResponse.graph_intellectual_entity?.[0]?.id) {
+		if (!ie?.id) {
```
reunefe:
```suggestion
		if (!ieObjectId) {
```

src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -1157,7 +1098,7 @@ export class IeObjectsService {
 
 		/* istanbul ignore next */
 		return compact(
-			await mapLimit(dbFiles, 20, async (includeFile): Promise<IeObjectFile> => {
+			await mapLimit(dbFiles || [], 20, async (includeFile): Promise<IeObjectFile> => {
```
reunefe: Waarom deze change? Lijn 1095 covered dit toch al?

reunefe (approved):

reunefe:
> should make it easier to get a database result from production and debug it locally in a test since the whole ie-object is fetched in a single query which should be more efficient than doing a promise.all of multiple smaller sub queries

Lijkt inderdaad eenvoudiger, al heb ik ergens de vrees dat dit zaken gaat breken gezien de drastische wijziging. Ook al zie ik geen issues met de code

PR #641, opened 2026-04-27
ARC-3639: increase performance of sending material requests

reunefe:


src/modules/material-requests/controllers/material-requests.controller.ts
```diff
@@ -274,11 +274,13 @@ export class MaterialRequestsController {
...
-				lastName: user.getLastName(),
-				language: user.getLanguage(),
-			});
+			this.materialRequestsService
+				.sendRequestList(materialRequests, sendRequestListDto, {
+					firstName: user.getFirstName(),
+					lastName: user.getLastName(),
+					language: user.getLanguage(),
+				})
+				.then(noop); // Let's not wait on the sending of the lists
```
bertyhell: add catch with console.error(new CustomError())

bertyhell (approved):

src/modules/material-requests/controllers/material-requests.controller.ts
reunefe: I actually did not add this because the method internally does all that, seems like overkill to add?

PR #646, opened 2026-04-30
feat: add herbruikbaarheid filter support in elasticsearch query builder 

RobbeBierebeeck:


src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
```diff
@@ -37,6 +37,7 @@ export enum IeObjectsSearchFilterField {
 	CONSULTABLE_ONLY_ON_LOCATION = 'isConsultableOnlyOnLocation',
 	CONSULTABLE_MEDIA = 'isConsultableMedia',
 	CONSULTABLE_PUBLIC_DOMAIN = 'isConsultablePublicDomain',
+	REUSABILITY = 'herbruikbaarheid',
```
bertyhell: prefer english
```suggestion
	REUSABILITY = 'reusability',
```

src/modules/ie-objects/elasticsearch/elasticsearch.consts.ts
```diff
@@ -195,6 +197,28 @@ export const DEFAULT_QUERY_TYPE: { [prop in IeObjectsSearchFilterField]: QueryTy
...
+		'https://creativecommons.org/licenses/by-sa/4.0/',
+		'https://creativecommons.org/licenses/by-nc/4.0/',
+	],
+	'misschien-herbruikbaar': [
+		'https://rightsstatements.org/page/UND/1.0/',
+		'https://rightsstatements.org/page/InC/1.0/',
+		'https://rightsstatements.org/page/CNE/1.0/',
+		'https://rightsstatements.org/page/InC-OW-EU/1.0/',
+		'https://rightsstatements.org/page/InC-RUU/1.0/',
+	],
```
bertyhell: prefer enum for categories, and use either the exact values from the database
eg:
* Vrij herbruikbaar
* Herbruikbaar onder voorwaarden
* Misschien herbruikbaar

, or english values for the enum + translate table for the database values

bertyhell (approved):

bertyhell (approved):

src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -185,11 +193,34 @@ export class IeObjectsService {
...
+			return [];
+		}
+
+		const response = await this.dataService.execute<
+			GetReusabilityRightsIrisQuery,
+			GetReusabilityRightsIrisQueryVariables
+		>(GetReusabilityRightsIrisDocument, { categoryIds });
+
+		return uniq(response.graph_rights.map((rights) => rights.intellectual_entity_id));
+	}
```
bertyhell, line 222: the reuse rights iris list will almost never change. so it will be more performant if we fetch all of them and cache it for 1 day. so we don't have to wait for this function on every query.

so change the query to:
GetAllReusabilityRightsIris
and make a cached variant: GetAllReusabilityRightsIrisCached
similar to https://github.com/viaacode/hetarchief-proxy/blob/a94d5eac649bdff24393fdfbf4d6c5a23e5eaf17/src/modules/ie-objects/services/ie-objects.service.ts#L1662

then use the cached variant of the function here

bertyhell (changes requested):

RobbeBierebeeck:
- [ ]

PR #652, opened 2026-05-07
add AV access test cases to limit-access-to-object

RobbeBierebeeck:
…details spec (ARC-3594)

### DIFF
--- a/src/modules/ie-objects/services/ie-objects.service.ts
+++ b/src/modules/ie-objects/services/ie-objects.service.ts
@@ -4,13 +4,20 @@ import { retry } from 'async';
 
 import { DataService, PlayerTicketService } from '@meemoo/admin-core-api';
 import { CACHE_MANAGER } from '@nestjs/cache-manager';
-import { Inject, Injectable, InternalServerErrorException, Logger, NotFoundException, } from '@nestjs/common';
+import {
+	Inject,
+	Injectable,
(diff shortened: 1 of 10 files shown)


src/modules/ie-objects/services/ie-objects.service.ts
```diff
@@ -676,9 +675,7 @@ export class IeObjectsService {
 			(licenses.includes(IeObjectLicense.PUBLIEK_CONTENT) ||
 				licenses.includes(IeObjectLicense.BEZOEKERTOOL_CONTENT));
 		const rightsInfo = shouldExposeRightsInfo
-			? ((await this.getRightsInfoByIeObjectIdFromDb(ieObjectId)) ?? {
-					reuseLabel: NO_RIGHTS_INFO_LABEL,
-				})
+			? await this.getRightsInfoByIeObjectIdFromDb(ieObjectId)
```
bertyhell, line 678: shouldn't we just get the info in the main query, so we don't have to do another round trip to the database?
i think we should be able to add it here: https://github.com/viaacode/hetarchief-proxy/blob/848cbd900bfef359e061b6bdb9a5b7ec45bd6cd9/src/modules/ie-objects/queries/getIeObjectDetail.graphql#L1

bertyhell (changes requested):

PR #660, opened 2026-05-19
ARC-3645: Add VAT and address to summary pdfs

reunefe:


src/modules/material-request-messages/services/material-request-pdf-generator.ts
```diff
@@ -336,6 +336,22 @@ export class MaterialRequestPdfGeneratorService {
...
+			this.greyText(doc, contentWidth, materialRequest.requesterOrganisationAddress);
+		}
+		if (
+			materialRequest.requesterOrganisationPostalCode ||
+			materialRequest.requesterOrganisationLocality
+		) {
+			this.greyText(
+				doc,
+				contentWidth,
+				`${materialRequest.requesterOrganisationPostalCode || ''} ${materialRequest.requesterOrganisationLocality || ''}`
```
bertyhell: use join, so we don't get a leading space if postalcode is undefined
```suggestion
				[materialRequest.requesterOrganisationPostalCode || '', materialRequest.requesterOrganisationLocality || ''].join(' ')
```

bertyhell (approved):

PR #671, opened 2026-06-01
ARC-3574: Send email when export fails

reunefe:


.env/.env.template
```diff
@@ -99,6 +99,7 @@ CAMPAIGN_MONITOR_TEMPLATE_MATERIAL_REQUEST_DOWNLOAD_EXPIRE_SOON__EN=
 CAMPAIGN_MONITOR_TEMPLATE_MATERIAL_REQUEST_DOWNLOAD_EXPIRE_SOON__NL=
 CAMPAIGN_MONITOR_TEMPLATE_MATERIAL_REQUEST_DOWNLOAD_DOWNLOADED__EN=
 CAMPAIGN_MONITOR_TEMPLATE_MATERIAL_REQUEST_DOWNLOAD_DOWNLOADED__NL=
+CAMPAIGN_MONITOR_TEMPLATE_MATERIAL_REQUEST_DOWNLOAD_FAILED__NL=
```
bertyhell, line 103: This one doesn't have an english version. So we should add a comment to explain that

```suggestion
# email to meemoo when an export jobs fails
CAMPAIGN_MONITOR_TEMPLATE_MATERIAL_REQUEST_DOWNLOAD_FAILED__NL=
```

bertyhell (approved):

PR #693, opened 2026-06-29
ARC-3591: add base mapping function for common request info

reunefe:


src/modules/campaign-monitor/services/campaign-monitor.service.ts
```diff
@@ -590,22 +593,14 @@ export class CampaignMonitorService implements OnApplicationBootstrap {
...
-					materialRequest.maintainerSlug
-				}/${materialRequest.objectSchemaIdentifier}`,
-				request_type: MATERIAL_REQUEST_TYPE_TRANSLATIONS[materialRequest.type],
-				request_description: materialRequest.reason,
 				request_url: stringifyUrl({
 					url: `${this.configService.get('CLIENT_HOST')}/${requestUrl}`,
 					query: {
 						materialRequest: materialRequest.id,
 					},
 				}),
```
bertyhell: shouldn't we include these fields (cp_name, request_url) into the mapMaterialRequest function
so we stay consistant for all emails?

bertyhell (approved):

src/modules/campaign-monitor/services/campaign-monitor.service.ts
reunefe: Yeah, maybe although the request_url is not always included but only for the requester since it goes to the outgoing request screen

PR #694, opened 2026-06-29
ARC-3747: align event insertion

reunefe:


src/modules/events/services/events.service.ts
```diff
@@ -24,6 +26,25 @@ export class EventsService {
 		});
 	}
 
+	public mapUserToEventData(user: SessionUserEntity | User | null): {
```
bertyhell: why accept both SessionUserEntity and User
shouldn't we just acceppt user and the code that calls this should do sessionUser.getUser()?

if not, i would try to generalize a bit like:
const simpleUser = (user as SessionUserEntity).getUser() ?? user;

So you can use simpleUser 3 times later on

bertyhell (approved):

src/modules/events/services/events.service.ts
reunefe: I didn't even know that method existed 😆  WIll change it indeed

PR #698, opened 2026-07-02
feat(ARC-3769): make one endp. to fetch theme & objects random | ordered

bertyhell:
https://meemoo.atlassian.net/browse/ARC-3769

orderDirection: 'asc' | 'desc' | 'random'

in the screenshot, the objects are not returned, because the seed contains QAS objects instead of localhost db ie-object ids
image


reunefe (approved):
Toch niet meer via slug/uuid via dezelfde endpoint?

